PR #55, opened 2015-01-14
Do not call AsyncTask.get() in UI thread

### LINKED ISSUE #54: Call AsyncTask.get in sendGetData and setPostData will block UI thread
In `sendGetData` and `setPostData` in `Emitter` [0], they call `AsyncTask.get` to get HTTP response. `sendGetData` and `setPostData` in `Emitter` are called in UI thread since they use `AsyncTasks`. However, they also use  `AsyncTask.get` [1] to get HTTP response, which will block the UI thread if HTTP response is not returned. It might slow down UI thread and cause bad UI experience.

[0] https://github.com/snowplow/snowplow-android-tracker/blob/master/snowplow-android-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/android/emitter/Emitter.java#L181-L217
[1] https://developer.android.com/reference/android/os/AsyncTask.html#get()

czchen:
This patch ensures UI thread will not be blocked by HTTP response.

Closed #54.


alexanderdean:
Closing in preference to the Rx tack we have taken in 0.3.0.

PR #91, opened 2015-02-13
[WIP] Minimizing footprint by removing Rx and RetroLamba

duncan:
This is currently a work in progress. It’s emitting events and we're collecting them with our collector and parsing them into our event stream with this, but it’s not yet fully ready to go. 

Obviously, if y'all decide to do a minimal Android snowplow collector, there's all sorts of questions about how to structure. New repo, etc?

There is one commit y'all should cherry pick ASAP involving double wrapping unstructured events: https://github.com/6wunderkinder/snowplow-android-tracker/commit/477b12a18457ad16bf74f4b9e8f019b57880e515 I haven't yet looked to see if this double wrapping effects/interacts with the other structured events, but it totally threw us off on the back end where we’re pulling the raw kinesis thrift events and processing a slice out into our existing data infra for a bit.


alexanderdean:
Hey @duncan! Many thanks for this, this looks cool. Look forward to brainstorming how we support both approaches...

duncan:
At a guess, there are two approaches—One is to create yet another project called android-common and have both a android-simple and android-rx project pull from it. The other is to bundle the rx stuff into a different target in this repo and then we sort out how to have Rx override the emitter and other bits. 

Let's hash it out and sort out how to proceed! But probably after your 0.3.0 release today :) We'll keep going with this in the meantime for our own usage and sort it out. 

Alles gut!

duncan:
Also, happy to jump on the phone/skype with you and @jbeemster if that helps! Just let me know. I very much want to do this in a way that’s a win-win-win for everyone.

duncan:
To come back around to the schema discussion, one thing sticks out. When you track an unstructured event from iOS, you can do so without defining your own schema. You can call `[tracker trackUnstructuredEvent:@{@"foo":@"bar"}]`, and get:

```
{"schema":"iglu:com.snowplowanalytics.snowplow/unstruct_event/jsonschema/1-0-0","data": {"foo": "bar"}}
```

When you do it from Android, you have to put in a schema into the call using the SelfDescribingJson object. When you do that with an unstructured event type, you get:

```
{"schema":"iglu:com.snowplowanalytics.snowplow/unstruct_event/jsonschema/1-0-0","data": { "schema":"iglu:com.snowplowanalytics.snowplow/unstruct_event/jsonschema/1-0-0","data": "foo": "bar", "parameters": {}}}}
```

So, the Android API looks like it's making a further enforcement beyond the iOS API that you have to define schemas on unstructured events.

alexanderdean:
Hey @duncan! Right, we got the iOS Tracker release out, huge thanks for your contributions to that.

Next step is the Android Tracker release - we are in code freeze on this and hope to get it out by tomorrow (again, customer commitment). After that, it would be great to have a phone/skype to brainstorm on how we can support your approach alongside the rx approach.

On the schema discussion:

> So, the Android API looks like it's making a further enforcement beyond the iOS API that you have to define schemas on unstructured events. 

You are right, the Android API is making this further enforcement. We will slowly move the other trackers to make this enforcement as well.

In your example - you wouldn't ever box an `unstruct_event` inside another one:

```
{"schema":"iglu:com.snowplowanalytics.snowplow/unstruct_event/jsonschema/1-0-0",
 "data": { 
  "schema":"iglu:com.snowplowanalytics.snowplow/unstruct_event/jsonschema/1-0-0",
  "data": "foo": "bar", "parameters": {}}
  }
}
```

Instead, there is an example of an unstructured event in the new Android Tracker documentation here:

https://github.com/snowplow/snowplow/wiki/Android-Tracker#46-track-unstructured-events-with-trackunstructuredevent

With that save_game event example, the ultimate payload would look like this:

```
ue_pr=
{"schema":"iglu:com.snowplowanalytics.snowplow/unstruct_event/jsonschema/1-0-0",
 "data": {
    "schema": "iglu:com.acme/save_game/jsonschema/1-0-0",
    "data": {
      "levelName": "Barrels o' Fun",
      "levelIndex": 23
    }
  }
}
```

The two levels of boxing can be a little confusing at first glance:

1. The first level is controlled by the tracker, and says, "the data payload should conform to a Snowplow `unstruct_event` v1-0-0"
2. The second level is controlled by the developer, and says, "the data payload should conform to an Acme `save_game` v1-0-0

The whole Snowplow pipeline is dependent on self-describing JSONs (/dictionaries/maps/hashes) to operate. I know it's confusing that trackers like the JavaScript and Objective-C still look like you can send in schemaless "unstructured" events, when in fact all events must be accompanied by a schema for the rest of the pipeline to function. In other words: 

> When you track an unstructured event from iOS, you can do so without defining your own schema

is a bug not a feature. Hopefully as we evolve the trackers' APIs and documentation to be more like the Android Tracker, this will make this need-for-schema clearer and prevent instrumentation mistakes causing data validation/loading issues downstream...

duncan:
Ok, that explains why the event processing is failing for me inside of some components for my own unstructured events, but not in others. Our first implementation strategy for Snowplow is that we’re using the trackers and collector into Kinesis, storing the raw Thrift events in S3, and tapping into the event stream with our own Kinesis app to push events into our existing analytics back end. That must have masked my understanding of the issue.

So, to make sure I have this right, in the ObjC client, where I’m currently calling:

```
[tracker trackUnstructuredEvent:@{@"event":@"foo", @"params": {@"foo": "bar"}}];
```

I really should be calling:

```
NSDictionary *event = @{@"event":@"foo", @"params": {@"foo": "bar"}};
NSString *schema = @"iglu:com.wunderlist/client_event/jsonschema/1-0-0";
[tracker trackUnstructuredEvent:@{@"schema":schema, @"data": event];
```

And, ideally, we'd have a registered schema somewhere...

duncan:
Addition: Thanks for taking the time to explain that to me. If I’ve got my understanding right, I'll adjust our ObjC client and our Kinesis -> SNS back end to do appropriate things. And, I'll take a look at pushing all he Snowplow events into their own Redshift table and taking a look at them in their native Snowplow format.

alexanderdean:
Hey @duncan - that's exactly right. And your pipeline makes sense as to why you weren't running into problems with the schemaless approach yet.

So yes, put that schema boxing around your events in the Objective-C Tracker. You can strip back to the final data node for your existing analytics back-end. To process the events in S3 through to Redshift you would need to setup your schema repo (Iglu), a hosted paths bucket and deploy the table(s) into Redshift.

For our ElasticSearch loading from Kinesis you don't need to create the Iglu repo - yet. Nor for the BigQuery loading that's coming soon...

alexanderdean:
(But eventually all our loads will be gated by schema checks so you will have to define that Iglu repo.) In any case, having a versioned repository of your event schemas is a big win.

duncan:
Ok, cool. Now I've got the picture. Now that I see it, I'm tempted to argue that the `trackUnstructuredEvent` method is misnamed and should be replaced with one that's a bit clearer since, with a schema, it is structured. It’s just not using a pre-defined structure like the other tracking methods. 

For the time being, I'll update our iOS/ObjC clients to emit the right payload, fix our stuff up, and then return to digging into a minimalistic Android lib that will work for our needs.

Thanks again for the info!

alexanderdean:
Hey @duncan - my pleasure, it's always good chatting about this stuff.

You are right, unstructured event is misnamed and unfortunate. We came up with it before we found out about JSON Schema, *and* we already had "structured events" (Google's action-category-value-etc). So it's kind of a legacy thing at this point.

Custom contexts are wrongly named too now - they can be any snapshotted entity, as per this [blog post](http://snowplowanalytics.com/blog/2015/01/18/modeling-events-through-entity-snapshotting/).

Rather than fixing piecemeal, we'll do a tracker overhaul at some point and update everything to our event grammar thinking, but that is a way off...

alexanderdean:
Hey @duncan - can you rebase this off of `feature/rx-free` now? (We have just merged in 0.3.0). Cheers!

duncan:
Sorry to let this sit a few days. I was caught up in a few other things... I'll be looking at this again today.

alexanderdean:
No worries!

duncan:
Closing out in favor of #103’s approach.

PR #103, opened 2015-02-24
[WIP] Split out Common module and add Lite tracker

duncan:
This is a proposal for structuring the project with two outputs: The existing `snowplow-android-tracker` with Rx and RetroLamba and a `snowplow-android-tracker-lite` without those components. Everything that’s common is in `snowplow-android-common`. 

This isn’t quite done as the `Emitter`, `EventStore`, and `Tracker` classes are the same in each product instead of being behind some sort of interface and builder. But it should be far enough to judge if doing a multiple target approach in the same repository is acceptable.


alexanderdean:
This is really cool Duncan! @jbeemster is reading through the PR carefully and will follow up with some detailed feedback this afternoon...

duncan:
Cool. @jbeemster: The other way we could do it is to have the Rx version extend the Lite version. Then we'd only have two targets. I was neutral between those two options, but went this way to make the common parts a bit more obvious.

jbeemster:
Hey @duncan.  I think having a core with two separate versions makes a lot of sense.  It will also allow us to potentially add different setups for the tracker with different styles of emitters as we need them.  So I really like this idea!

Really all that is left to be done is, as you said, to move the last chunks of code into common and replace them with interfaces/builders for the different implementations.

A couple of thoughts though:
* Should we call the Rx library `snowplow-android-tracker-rx` to keep in sync with the naming for the `lite` version?
* The `build.gradle` file for the `snowplow-android-core` should probably be the only one to contain the function to generate the `.../generated/Version.java` file.  Will also need to update the root gitignore file to clean it up.

Other than that it all looks good to me!

alexanderdean:
Hey guys - agree that `-rx` would make sense as an suffix. I suspect treating the Rx version as the "default" tracker would come back to bite us sooner or later...

duncan:
Cool! Thanks! :) I agree on the `-rx` suffix and doing the generated Version file in one place. I'll get that implemented in the next round of commits

snowplow-android-core/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+        targetSdkVersion 21
+        versionCode 1
+        versionName "1.0"
+    }
+    buildTypes {
+        debug {
+            buildConfigField "String", "TRACKER_LABEL", "\"andr-$project.version\""
+            buildConfigField "String", "TRACKER_VERSION", "\"$project.version\""
+        }
+        release {
```
duncan: @jbeemster I found that I could do this instead of generating a Version.java and thought it was a nice way to take that out. However, there may be another reason for Version.java that I'm not aware of and we should keep it. Thoughts?

jbeemster: Hey @duncan, as far as I am aware we only need those vars available so that we can pass them through with events.  So as long as this still makes the information available to the Tracker I don't see an issue.

alexanderdean: Sounds good to me!

duncan:
Ok, this is pretty far along at this point. Everything builds smoothly, most tests pass, and the restructuring should pretty much be there. At this point, it's getting pretty hard to see the details in the diffs and it's probably easier to just scope out the branch: https://github.com/wunderlist/snowplow-android-tracker/tree/commonify

Things left to do: 
[] Set up/verify jars for distribution builds
[] Fix all tests back up
[] Actually run each client variant in a real setup to test events are being sent

More to come...

alexanderdean:
Awesome!

duncan:
Ok, these last few commits put in tests for the JSON generator tracker payloads and then uses those tests to verify a clean removal of Jackson as a dependency. That's big set of methods that won't take up DEX space and makes it so that we can embed this into our Android app. 

As of today, we'll be working on doing just that and shaking out any remaining issues from the dependency reduction. So, as of now, my checklist for this is:

-[ ] set up and verify jars from build
-[ ] embed into our application
-[ ] generate and verify events to our collectors

Wheee! I know this is a huge branch/PR now. I think I'll be ready to call it good today or over the weekend and then submit for feedback. :)

alexanderdean:
Sweet! Great work @duncan - we really appreciate all your efforts. It's going to be great to offer Androiders two different Snowplow tracker flavors.

jbeemster:
Hi @duncan!  Was just wondering if there was any news on the Lite Android tracker implementation?  We would love to get this integrated and out into the wild as an extra tracker option.

duncan:
Hi! Sorry, I was out last week at TED. We've got this out and deployed in our latest Android releases and all seems well. Yes, let's get this merged up and integrated!

duncan:
Ok, our tracker has been deployed and sending events for quite some time now and I think it's ready to integrate and get out there. Let's do it!

alexanderdean:
Awesome! Assigning to @jbeemster for 0.4.0...

duncan:
What can we do on our end? I can take another top to bottom review cycle early this week if it'd help?

alexanderdean:
A top-to-bottom review would be most excellent @duncan - thanks so much!

alexanderdean:
This is done, thanks so much!

PR #186, opened 2016-02-25
hotfix/0.5.4

jbeemster:


CHANGELOG
```diff
@@ -1,3 +1,8 @@
+Version 0.5.4 (2016-02-25)
+--------------------------
+Core: fixed input string \"null\" is not a valid UUID with previousSessionId on Android Versions lower than KitKat (#185)
```
alexanderdean: Change to imperative from past tense

snowplow-android-core/src/main/java/com/snowplowanalytics/snowplow/tracker/utils/Util.java
```diff
@@ -114,14 +114,7 @@ private static Object getJsonSafeObject(Object o) {
...
         } else if (o == null) {
-            return new Object() {
-                @Override
-                public boolean equals(Object o) { return o == this || o == null;  }
-                @Override
-                public String toString() {
-                    return "null";
-                }
-            };
+            return JSONObject.NULL;
```
alexanderdean: Looks much safer

bernardosrulzon:
@jbeemster Looks great! I think it's worth reminding users on the release announcement that they'll need (1) to update the schema to allow the `null` string until they have an updated version on Google Play, and (2) re-process past events that ended up in the bad bucket. 

Btw - I've updated the readme of the Hadoop Bad Rows job to make it work with the AWS CLI on snowplow/#2473

Does that make sense?

PR #217, opened 2016-08-16
Release/0.6.0

jbeemster:


jbeemster:
@alexanderdean have ensured that the application crash will be truncated at the maximums specified in the jsonschema.

PR #242, opened 2017-01-13
Hotfix/0.6.2

jbeemster:


CHANGELOG
```diff
@@ -1,3 +1,13 @@
+Version 0.6.2 (2017-01-13)
+--------------------------
+Ensure we do not pass invalid information to application_error event (#237)
+Mobile Context: offline events not being recorded as such (#239)
+GeoLocation context: changes in position not being reflected (#240)
+Bump Android Gradle to 2.2.3 (#241)
+Logger crash when the value contains '%' (#238)
+Update Copyright to 2015-2017 (#243)
+Ensure Vagrant environment can build and test Tracker (#227)
+
```
alexanderdean: Josh can we do another pass on these ticket names? They are a bit hit-and-miss - inconsistent capitalization, mixture of formats, and some are descriptive rather than actionable (e.g. change `Logger crash ...` to `Fix logger crash ...`)

jbeemster: Sure!

Fix logger crashing when a value contains '%'
Update copyright to 2015-2017
Ensure vagrant environment can build and test tracker

Is that what you are after?

alexanderdean: Sounds good, plus "Mobile context:" not "Mobile Context:"

jbeemster:
Okay @alexanderdean I have cleaned up the CHANGELOG.

I have built the snowplow-demo-app on my phone and everything appears to be working as expected so I think we are good to go unless you can think of anything else?

alexanderdean:
LGTM!

PR #303, opened 2019-04-16
Release 1.1.0

oguzhanunlu:
- application context
- application install
- updated screenview with auto tracking & screen context


oguzhanunlu:
Thanks for the feedback @mhadam ! `editor.apply()` is an async function but other shared preferences works synchronously, I'll keep the futureCallable and remove file ops.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/events/ScreenView.java
```diff
@@ -19,15 +19,27 @@
 import com.snowplowanalytics.snowplow.tracker.payload.SelfDescribingJson;
 import com.snowplowanalytics.snowplow.tracker.payload.TrackerPayload;
 
+import java.lang.reflect.Parameter;
```
mhadam: Is this used?

snowplow-demo-app/src/androidTest/java/com/snowplowanalytics/snowplowtrackerdemo/context/ApplicationContextTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    private Context context;
+
+    @Before
+    public void setUp() {
+        context = InstrumentationRegistry.getInstrumentation().getTargetContext();
+    }
+
+    @Test
+    public void testGetApplicationContext() {
```
mhadam: Can we call the function itself instead of rewriting it here?

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/events/ScreenView.java
oguzhanunlu: Nope, will delete it.

snowplow-demo-app/src/androidTest/java/com/snowplowanalytics/snowplowtrackerdemo/context/ApplicationContextTest.java
oguzhanunlu: It looks like a rewrite but `getApplicationContext` is called on line 57. We need to fetch `versionName` and `versionCode` so that we can compare result of `getApplicationContext` with the fetched values. Do I miss something?

PR #308, opened 2019-05-20
Add support for global contexts

oguzhanunlu:


oguzhanunlu:
thanks for the feedback @mhadam , closing the PR since I already cherry-picked my work into release branch, I'll address your feedback there

PR #321, opened 2019-07-23
Release/1.3.0

oguzhanunlu:
* Fix database cursor usage (#330)
* Comply with isLimitAdTrackingEnabled (#329)
* Integrate Snowplow Micro (#318)
* Trim travis setup (#324)
* Remove vagrant setup (#317)


snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/Tracker.java
```diff
@@ -957,4 +964,27 @@ public void removeGlobalContext(String tag) {
...
+    private String basisForProcessing;
+    private String documentId;
+    private String documentVersion;
+    private String documentDescription;
+
+    /**
+     * Enables gdpr context to be sent with every event
+     */
+    public void enableGDPRContext(BASIS basisForProcessing, String documentId, String documentVersion, String documentDescription) {
+        String basis = basisForProcessing.toString().toLowerCase();
```
colmsnowplow: Since this is an [enum](https://github.com/snowplow/iglu-central/blob/248b3be05b097693363e014b902008fa6cb3bfc8/schemas/com.snowplowanalytics.snowplow/gdpr/jsonschema/1-0-0#L14) in the schema, we've opted to [make it an enum in the javascript tracker](https://github.com/snowplow/snowplow-javascript-tracker/blob/a2a12a5c1a2065db8df250eb1bf6e31900e81a0c/src/js/tracker.js#L111).

Not sure how practicable that is for Android

oguzhanunlu:
Thanks @colmsnowplow ! I've extended the PR after your review by 1 commit, could you please review that last commit too?

.travis.yml
```diff
@@ -9,54 +9,38 @@ dist: trusty
 
 env:
   global:
-    - EMULATOR_API=24
+    - EMULATOR_API=16
```
jbeemster: How come we are dropping the API back so much?

.travis.yml
```diff
@@ -9,54 +9,38 @@ dist: trusty
...
-  - yes | sdkmanager "platforms;android-"$ANDROID_API
-  - yes | sdkmanager "emulator"
-  - yes | sdkmanager "extras;android;m2repository"
-  - yes | sdkmanager "extras;google;m2repository"
-  - yes | sdkmanager --update
-  - yes | sdkmanager --licenses
-  - sdkmanager --list --verbose || true
+#  - docker pull snowplow/snowplow-micro:0.1.0
+#  - wget https://bin.equinox.io/c/4VmDzA7iaHb/ngrok-stable-linux-amd64.zip
+#  - unzip ./ngrok-stable-linux-amd64.zip
```
jbeemster: If these are commented out do we need them?

.travis.yml
```diff
@@ -65,6 +49,10 @@ script:
   # travis builds timeout after 10 min of not seeing any output, hence the below work-around
   # default travis_wait method doesn't show output until command finishes while this way allows us to see it
   - (while sleep 9m; do echo "=====[ $SECONDS seconds still running ]====="; done &)
+#  - docker run -d --mount type=bind,source=$TRAVIS_BUILD_DIR/micro_config,destination=/config -p 9090:9090 snowplow/snowplow-micro:0.1.0 --collector-config /config/micro.conf --iglu /config/iglu.json
+#  - docker ps -a
+#  - echo $MICRO_SUBDOMAIN
+#  - ./ngrok http -authtoken $NGROK_AUTH -subdomain $MICRO_SUBDOMAIN -log=stdout 9090 > /dev/null &
```
jbeemster: Same here

snowplow-demo-app/build.gradle
```diff
@@ -38,8 +56,11 @@ dependencies {
...
     // test
-    androidTestImplementation 'androidx.test:runner:1.1.0'
-    androidTestImplementation 'androidx.test:rules:1.1.0'
-    androidTestImplementation 'androidx.test.ext:junit:1.0.0'
-}
+    androidTestImplementation 'androidx.test:runner:1.2.0'
+    androidTestImplementation 'androidx.test:rules:1.2.0'
+    androidTestImplementation 'androidx.test.ext:junit:1.1.1'
+    androidTestImplementation 'androidx.test.espresso:espresso-core:3.2.0'
+}
```
jbeemster: Nit newline

snowplow-demo-app/src/androidTest/java/com/snowplowanalytics/snowplowtrackerdemo/uitest/UITest.java
```diff
@@ -0,0 +1,179 @@
+package com.snowplowanalytics.snowplowtrackerdemo.uitest;
```
jbeemster: Nit header missing

snowplow-demo-app/src/androidTest/java/com/snowplowanalytics/snowplowtrackerdemo/uitest/UITest.java
```diff
@@ -0,0 +1,179 @@
...
+    public void t1_sendDemoEvents() throws InterruptedException, IOException, JSONException {
+        client.newCall(resetRequest).execute();
+
+        Espresso.closeSoftKeyboard();
+        onView(withId(R.id.emitter_uri_field)).perform(replaceText(micro_uri), closeSoftKeyboard());
+        onView(withId(R.id.emitter_uri_field)).check(matches(withText(micro_uri)));
+        onView(withId(R.id.btn_lite_start)).perform(click());
+
+        // TODO this is not best practice for idling resources
+        Thread.sleep(20000);
```
jbeemster: This seems excessive and is introducing 20 seconds to a test suite!  This should really be refactored to not require any sleeps.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/Tracker.java
```diff
@@ -957,4 +964,27 @@ public void removeGlobalContext(String tag) {
...
+    private boolean gdprContext = false;
+    private String basisForProcessing;
+    private String documentId;
+    private String documentVersion;
+    private String documentDescription;
+
+    /**
+     * Enables gdpr context to be sent with every event
+     */
+    public void enableGDPRContext(BASIS basisForProcessing, String documentId, String documentVersion, String documentDescription) {
```
jbeemster: This function seems like it should only set the global values if `gdprContext = false`.  Otherwise if you are mid sending events when it is on and switch it off you might get a thread race condition.

Also worth making this function thread safe with tracking calls to ensure you cannot have any race conditions between these tracker variables.

Atomic booleans could be useful here!

.travis.yml
oguzhanunlu: The tracker supports SDK 14 at the minimum and emulator startup time of 16 is lower than 24 in Travis. It reduced build time from ~30 min to ~22-23 min. Do you think having a recent version will bring an advantage?

.travis.yml
oguzhanunlu: Micro integration is half complete, i.e. integration tests are run locally only, not in Travis, and I opened #328 to address the other half. These lines are commented out because they aren't necessary until 328 is done. Would you prefer to save this incomplete work somewhere else and delete from here? Or anything else?

snowplow-demo-app/src/androidTest/java/com/snowplowanalytics/snowplowtrackerdemo/uitest/UITest.java
oguzhanunlu: It was based on demo app, shorter sleeps can end the tests without letting the app send all events.

I'm aware of the disadvantage and I was planning to change the approach later since this isn't a proper way. Although test duration won't change much, I'll change the approach right in this release.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/Tracker.java
oguzhanunlu: > This function seems like it should only set the global values if gdprContext = false. 

It mimics how our js tracker offers gdpr context. Are you sure on this? 

Thanks for catching thread-safety. I'll make the method `synchronized`.

.travis.yml
jbeemster: So API version 16 is used only by 1.2% of devices according to this page:

> https://developer.android.com/about/dashboards

I would argue that using a later version is more useful in that you are testing on more relevant releases.

.travis.yml
jbeemster: Maybe add a comment that these commented lines are related to that ticket and leave them in then?

snowplow-demo-app/src/androidTest/java/com/snowplowanalytics/snowplowtrackerdemo/uitest/UITest.java
jbeemster: I think for this kind of thing its better not to leave it!  Let me know if you need any help solving it.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/Tracker.java
jbeemster: Mimic the same behaviour then but thread safety is really important for Tracker functions!

.travis.yml
oguzhanunlu: I see your point, I'll revert it 👍

oguzhanunlu:
Hey @jbeemster , I addressed all your feedback, especially on `thread.sleep` usage in UI test and thread safety of gdpr context mechanisms, however the scope of the release has to be cut (configurable emitter post path and gdpr context is postponed) and I need to ask for review again on updated commits. Thanks in advance.

Issue #314, opened 2019-07-08
Fix sending duplicate contexts


oguzhanunlu:
`1.2.0` wasn't successful on leveraging global contexts.

SzotsLevente:
@oguzhanunlu also please check custom context, it seems that it got broke in the process as well

oguzhanunlu:
@SzotsLevente , do you refer to global contexts or plain custom contexts that we attach to events one by one? Do you mind sharing more details on what got broke in which process etc. Thanks!

SzotsLevente:
custom context i mean by the set of data that is constructed using hashmap

djohannot:
global context are broken, tou add to them without clearing anytime, so it never stop to grows. Also having global context that is modified by a third party thread (the executor) makes it veeeeeery dangerous and unreliable i think, you should check your code and add more unit test that are more aggressive with your implementation (do loop of event sending and add network delay, check the size of your global context growing, etc.)

oguzhanunlu:
Hi @djohannot , thanks for the feedback, we are trying to improve our test suite. Could you provide us with more details on your global contexts usage? We'd like to see what is broken and fix it within the next release. Thanks in advance.

YSDC:
Problem is the tracker contexts are added indefinitely (Mobile, location, session contexts etc). In our case we have a look of events created and as the global contexts are never cleared it keeps continue added them

YSDC:
So even if I clear the global contexts before calling track it adds hundred of them

YSDC:
In fact I don’t really see the gain of having the global contexts added in a list in each event. Should be a map to avoid adding duplicates. Don’t you think?

oguzhanunlu:
Hi @YSDC , thanks for the feedback! Currently the tracker doesn't check if there are duplicate global contexts. We haven't introduced any safeguard in the initial release but will be improving the way global contexts are used.

Regarding that tracker adding multiple/hundreds of the same context, are you sure there is no custom context added to events manually in your SDK usage, or calling `addGlobalContext` in a loop etc.?

We can solve this issue a lot faster if we know how it happened. Do you mind sharing more details on your SDK usage so that we can reproduce the case and fix it? Thanks in advance.

djohannot:
You are welcome. I add custom contexts to my events, but these are not the one that are duplicated. It's the one defined by snowplow. 

[...]

PR #322, opened 2019-07-23
Bugfix/duplicate globalcontexts

djohannot:
Replace the list of global context with a map to avoid duplicates


oguzhanunlu:
Hi @djohannot , thanks for the PR and sorry for the late response. I was occupied with other work.

I think I detected the issue clearly and explained [here](https://github.com/snowplow/snowplow-android-tracker/issues/314#issuecomment-516409106). Fix is [here](https://github.com/snowplow/snowplow-android-tracker/pull/323). What do you think?

Nevertheless, we still need to handle duplication case since there is no mechanism in place to prevent a user from adding same context multiple times. I'd like to incorporate your work to the next release `1.3.0` after you sign the CLA.

AlexBenny:
Closing as it's a fix for an old bug already fixed. Also this implementation will be obsolete with the upcoming tracker v.2.0 as it comes with a slightly different implementation for the Global Contexts feature.

Issue #336, opened 2019-10-11
Fix ScreenState incompatibility with Screen schema


oguzhanunlu:
`ScreenState` is the data structure behind the scenes we use to create both Screen View events and Screen Context. It offers a constructor without arguments (creating `id` field only) which isn't compatible with [the updated screen schema](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.mobile/screen_view/jsonschema/1-0-0) which requires both `name` and `id` fields at least.

chuwy:
@oguzhanunlu please don't forget to assign a milestone.

oguzhanunlu:
Thanks @chuwy , I forgot to do that.

teffi:
@oguzhanunlu Thanks for the fix. I noticed ScreenState's supposedly automatic name `getAutomaticName`  returns null unless you call `updateWithFragment` or `updateWithActivity `. Was that an intentional design?

I also noticed that the iOS and Android tracker has different implementations of ScreenView. On Android id must be set (this wasn’t clear on the documentation) or else ScreenView event won’t send meanwhile iOS automatically creates an id.

oguzhanunlu:
Hi @teffi , considering latest version, [`getAutomaticName`](https://github.com/snowplow/snowplow-android-tracker/blob/master/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/tracker/ScreenState.java#L104-L114) doesn't return `null` even if you don't use `updateWithFragment` or `updateWithActivity`. It returns `"Unknown"` in case `getActivityField ` or `getFragmentField ` returns `null`, i.e. activity or fragment information isn't available. It was intentional so we always have a name.

You're right that both trackers differ in a good number of points including that. Unification of user experience across our tracker suite is one of our targets. Thanks for the notice.

Please let us know if we're not on the same page regarding `getAutomaticName`.

teffi:
Hi @oguzhanunlu I'm sorry.  Yes you're right. `getAutomaticName` now returns `"Unknown"`. I guess I just find the name confusing. `getAutomaticName` not being "automatic" at all since there's an assignment that needs to be done for it to work.  The lack of  `Fragment` or `Activity` constructor parameters in `ScreenState` or properties on `ScreenView` Builder which IMO sort of hides the dependency, an indication of the need to set certain values for another property to return the correct values.

oguzhanunlu:
[...]

PR #337, opened 2019-10-11
Release/1.3.2

oguzhanunlu:
See #336


colmsnowplow:
👍 LGTM. One question - what happens when a user sets their own custom screen name in the screen view event? Does `getAutomaticName()` take that one or does it always default to whatever it gets from the viewController?

oguzhanunlu:
Thanks Colm. On your question, when a user sets their own custom screen name in a screen view event, it isn't overwritten, it stays as is. By default, `getAutomaticName` fetches name of the current activity, equivalent of `viewController` from iOS, and that value is used until activity is changed.

colmsnowplow:
👍 Forgot to add the green tick yesterday. Done now apologies!

PR #344, opened 2020-01-07
Add tracker's error reporting (close #343)

### LINKED ISSUE #343: Add tracker's error reporting
The app where the tracker running on needs to be informed of any error that can affect the event tracking.
They might not compromise the event tracking but they are important to handle edge cases and to provide useful informations to support.

AlexBenny:
This feature is useful to track errors raised by the tracker to the collector as `diagnostic_error` events.
Previously the tracker treated internal errors managing them silently. They were visible in the logs of the app available on debugging only.
This can be particularly useful when an issue is not reproducible in debug mode but is visible on production. The user can collect and filter diagnostic_error events and report them to the support team if needed.

**Here the proposed documentation to review:**

---

### Tracker's diagnostic

The tracker is constantly tested with different platforms and versions but there are rare situations where an error can be reported. The tracker can manage the situation avoiding crashes of the app and assuring that the events are not lost. However, it can report errors to the collector as `diagnostic_error` events.

To activate this feature you only need to enable `trackerDiagnostic`:

```
TrackerBuilder trackerBuilder =
    new TrackerBuilder(emitter, namespace, appId, appContext)
        .trackerDiagnostic(true)
        ...
        .build();
Tracker.init(trackerBuilder);
```


snowplow-demo-app/src/main/java/com/snowplowanalytics/snowplowtrackerdemo/ErrorLogger.java
```diff
@@ -0,0 +1,14 @@
...
+
+import android.util.Log;
+
+import com.snowplowanalytics.snowplow.tracker.ErrorLogging;
+
+public class ErrorLogger implements ErrorLogging {
+    @Override
+    public void log(String source, String errorMessage) {
+        if (source.equals("Tracker") || source.equals("Session") || source.equals("ProcessObserver")) {
+            Log.e("Error reported >>> " + source, errorMessage);
```
paulboocock: Maybe worth adding a Tracking example here, send the error message into the collector? Just a thought but not a blocker.

AlexBenny:
@paulboocock I've started working on an example in the demo app with the purpose to use the tracker to track back possible errors as `application_error` as you suggested. But it's a bit dangerous because the previous implementation may generate infinite loops. An event can generate an error that generate another event and so on.

At the moment I've removed the risk allowing error tracking only for specific errors and excluding to add contexts (especially the Session one) to error events not important for our purpose.

Now the user app doesn't need to implement any delegate. The tracker is able to send error events autonomously. The user needs only to activate the flag `trackerDiagnostic` on setup of the Tracker.

Please, could you to confirm that I've correctly implemented the payload for the TrackerError events and sending it correctly? I've done some tests and it seems ok to me but I'd like to have your double check. That should be use to spot issues in the tracker along (and in sync) with the stream of events collected.

snowplow-demo-app/src/main/java/com/snowplowanalytics/snowplowtrackerdemo/Demo.java
```diff
@@ -315,6 +317,7 @@ private void initAndroidTracker() {
                 .mobileContext(true)
                 .geoLocationContext(true)
                 .applicationCrash(true)
+                .trackerDiagnostic(true)
```
paulboocock: I like this, much cleaner.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/ErrorLogging.java
```diff
@@ -0,0 +1,14 @@
+package com.snowplowanalytics.snowplow.tracker;
+
+public interface ErrorLogging {
```
paulboocock: Should this be called DiagnosticLogger? I'm not sure it is clear if this is related ApplicationCrash or TrackerDianogistic logging.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/events/TrackerError.java
```diff
@@ -0,0 +1,99 @@
...
+        Preconditions.checkNotNull(builder.message);
+
+        this.source = builder.source;
+        this.message = builder.message;
+        this.throwable = builder.throwable;
+    }
+
+    @Override
+    public Payload getPayload() {
+        return new SelfDescribingJson(TrackerConstants.APPLICATION_ERROR_SCHEMA, getData());
```
paulboocock: I'm in two minds as to whether this is the correct schema for this or if we should create a new one.
The application_crash schema was created to report when applications that contain the tracker throw unhandled exceptions, however in this case we are allowing customers to turn on diagnostics that allow the tracker to emit log events - the application hasn't crashed at all and this could inflate users crash statistics if they are counting this event type.
I wonder if a new iglu central schema would be better and more descriptive here, perhaps `tracker_log` or `tracker_diagnostic`?

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/tracker/ProcessObserver.java
```diff
@@ -137,7 +137,7 @@ public static void onEnterBackground() {
                     }
                 }
             } catch (Exception e) {
-                Logger.e(TAG, e.getMessage());
+                Logger.e(TAG, "Method onEnterBackground raised an exception: %s", e);
```
paulboocock: I'm just picking on this `Logger.e` as the message has changed in this PR but should we change all `Logger.e` to `Logger.track`?

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/Tracker.java
```diff
@@ -568,6 +605,40 @@ public void run() {
 
     // --- Helpers
 
+    private void addServiceEventPayload(Payload payload, List<SelfDescribingJson> contexts, String eventId) {
```
paulboocock: Is a ServiceEvent an event from within the tracker that doesn't include any Subject information?
A method comment like addEventPayload would be useful to desribe how this differs from addEventPayload would be great.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/ErrorLogging.java
AlexBenny: Changed to DiagnosticLogger

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/events/TrackerError.java
AlexBenny: I would go with `tracker_diagnostic` 👍

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/tracker/ProcessObserver.java
AlexBenny: As anticipated (sadly, weeks ago) it's not possible at the moment because the tracker doesn't allow service event and it can cause of infinite loops.
For now it's better limiting the error reporting for just few of them.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/Tracker.java
AlexBenny: Good idea!

AlexBenny:
@paulboocock  I've updated the code in order to track them internally rather than send them to the app.
It would be useful to send them to the app and to the collector, but for brevity (this PR really needs to be completed) I just report them to the collector and the user can see error in the logs.
I've also updated the documentation in the description of this PR. 

The associated schema is here: https://github.com/snowplow/iglu-central/pull/1023

PR #346, opened 2020-01-30
Session userId renewed for the same user (close #345)

### LINKED ISSUE #345: Session userId renewed for the same user
The session context is renewed on each app restart but the session userId should be persisted among relaunches of the app, unless the user delete and reinstall the app.
We noticed cases where the userId is renewed without app reinstall.

AlexBenny:
- Removed session data persisted on internal storage because it could cause exceptions forcing renewing of userId
- Session as singleton in order to avoid concurrent accesses on persisted session values


AlexBenny:
Thanks @paulboocock 
That's a good point. SQLITE is an incorrect information but it worked in that way so far. I prefer to fix it in the next minor or during the tracker refactoring.
I'm adding two tickets for that on Android and iOS repos.

PR #351, opened 2020-01-31
Fix screen context not correctly updated on screenView events (close #350)

### LINKED ISSUE #350: Fix screen context not correctly updated on screenView events
The screen context links events back to the specific screen view that generated the event.
At the moment it works correctly with the automatic screenView events only.
A screenView event handled outside the tracker is not able to update the internal tracker state causing the screen context not being updated correctly.

AlexBenny:
Fix #350 
As reported in the linked issue, the screen context wasn't correctly updated when the app send a screenView event. It used to work correctly only with automatic screenView tracking.
The bug is related to the software design so I needed to refactor a bit the logic behind ScreenState updating.

### DIFF
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/Tracker.java
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/Tracker.java
@@ -485,6 +485,11 @@ public class Tracker {
             return;
         }
 
+        if (ScreenView.class.isInstance(event) && screenState != null) {
+            ScreenView screenView = (ScreenView) event;
+            screenView.updateScreenState(screenState);
+        }
+
         Executor.execute(new Runnable() {
             @Override
             public void run() {
@@ -888,7 +893,9 @@ public class Tracker {
 
     /**
      * Track a screen with a screen state (many options)
+     * @deprecated Use track(Event) method passing a ScreenView event.
      */
+    @Deprecated
     public void trackScreen(ScreenState screenState) {
         this.screenState = screenState;
         SelfDescribingJson data = screenState.getScreenViewEventJson();
@@ -900,9 +907,11 @@ public class Tracker {
 
     /**
      * Track a screen only by name
+     * @deprecated Use track(Event) method passing a ScreenView event.
      */
+    @Deprecated
     public void trackScreen(String name) {
-        screenState.newScreenState(name, null, null);
+        screenState.updateScreenState(null, name, null, null);
         SelfDescribingJson data = screenState.getScreenViewEventJson();
         track(SelfDescribing.builder()
                 .eventData(data)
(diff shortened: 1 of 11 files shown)


snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/Tracker.java
```diff
@@ -485,6 +485,11 @@ public void track(final Event event) {
             return;
         }
 
+        if (ScreenView.class.isInstance(event) && screenState != null) {
```
paulboocock: Not for this PR but just a comment that when the refactor is underway, I'd love to see all this type checking removed. Makes me a little sad everytime I stumble across it.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/Tracker.java
```diff
@@ -888,7 +893,9 @@ public ScreenState getScreenState() {
 
     /**
      * Track a screen with a screen state (many options)
+     * @deprecated Use track method passing a ScreenView event.
```
paulboocock: I think might be a little clearer as: Use track(Event) method passing a ScreenView event

I only suggest this as I went looking for trackScreen(ScreenView) first.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/Tracker.java
```diff
@@ -900,9 +907,11 @@ public void trackScreen(ScreenState screenState) {
 
     /**
      * Track a screen only by name
+     * @deprecated Use track method passing a ScreenView event.
```
paulboocock: Potentially same here

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/events/ScreenView.java
```diff
@@ -159,4 +242,28 @@ public TrackerPayload getData() {
...
+
+    private static String getSnowplowScreenId(Activity activity) {
+        try {
+            Field field = activity.getClass().getField("snowplowScreenId");
+            Object reflectedValue = field.get(activity);
+            if (reflectedValue instanceof String) {
+                return (String)reflectedValue;
+            }
+        } catch (Exception e) {
+            // log here
```
paulboocock: Worth using the old logging system here to output to logcat? 
Seems likely that getField might throw a NoSuchFieldException and would be good for developers to know.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/Tracker.java
AlexBenny: Yeah same feeling for me!

AlexBenny:
Thanks Paul! Yep it will be a patch release. I've initially planned to put the fix in the 1.4.0 but it has been delayed too much so I'll push this in a different release branch (1.3.5) already prepared just for this fix.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/events/ScreenView.java
AlexBenny: Right, it's just a silent fault now.

PR #354, opened 2020-02-10
Cannot unset user id from subject (close #353)

### LINKED ISSUE #353: Cannot unset user id from subject
I'm currently setting the user id on a subject to `null` when the user logs out, however, instead of removing that value from the map, it sends `uid=null`. What is the recommended way to clear the `uid` from the subject map of values? 

```
Tracker.instance().subject.setUserId(user?.id)
```

AlexBenny:
The payload doesn't remove entries when the value is set to null. 
The expected behaviour (already implemented on ObjC tracker) is the entry removing when the value is null.

### DIFF
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/payload/TrackerPayload.java
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/payload/TrackerPayload.java
@@ -31,7 +31,8 @@ public class TrackerPayload implements Payload {
     @Override
     public void add(String key, String value) {
         if (value == null || value.isEmpty()) {
-            Logger.v(TAG, "The keys value is empty, returning without adding key: %s", key);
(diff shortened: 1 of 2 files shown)


snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/payload/TrackerPayload.java
```diff
@@ -31,7 +31,8 @@
     @Override
     public void add(String key, String value) {
         if (value == null || value.isEmpty()) {
-            Logger.v(TAG, "The keys value is empty, returning without adding key: %s", key);
+            Logger.v(TAG, "The keys value is empty, removing the key: %s", key);
```
paulboocock: > The keys value is empty, removing the key

I misunderstood this as the key was empty, not the associated value was empty.
I think this would be clearer given the names of the parameters:

> The value is empty, removing the key

Issue #377, opened 2020-05-22
Fix crash caused by InstallTracker component


AlexBenny:
InstallTracker send events just after the app installation.
If the tracker instance is dismissed rapidly it can cause the crash of the app as InstallTracker component tries to send an event using the tracker instance that no longer exist.

Issue #376, opened 2020-05-22
Fix race condition on tests


AlexBenny:
Lot of failures on testTrackWithNoContext where the MockWebServer receive an event generated by a different test.
The issue seems caused by zombie threads running from a previous test.

Issue #375, opened 2020-05-20
Move from Travis CI to GitHub Actions


AlexBenny:
We faced some issues building the Android tracker with the Travis script, especially for what concern the emulator and the API level.
GH Actions offers a marketplace where we can get third party modules ready to use with a minimal configuration.

Issue #373, opened 2020-04-30
Version 1.4.1 uses AndroidX instead of Support library


yonghow:
Hi,

My app is still using android.support library and it is working with version 1.2.1

I upgraded to version 1.4.1 and it throws class not found exception for `Landroidx/lifecycle/ProcessLifecycleOwner;`

Do we have to migrate to androidx library when using version 1.4.1?

```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: sg.trackerdemoapp, PID: 4942
    java.lang.NoClassDefFoundError: Failed resolution of: Landroidx/lifecycle/ProcessLifecycleOwner;
        at com.snowplowanalytics.snowplow.tracker.Tracker$1.run(Tracker.java:497)
        at android.os.Handler.handleCallback(Handler.java:739)
        at android.os.Handler.dispatchMessage(Handler.java:95)
        at android.os.Looper.loop(Looper.java:135)
        at android.app.ActivityThread.main(ActivityThread.java:5254)
        at java.lang.reflect.Method.invoke(Native Method)
        at java.lang.reflect.Method.invoke(Method.java:372)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)
     Caused by: java.lang.ClassNotFoundException: Didn't find class "androidx.lifecycle.ProcessLifecycleOwner" on path: DexPathList[[zip file "/data/app/sg.trackerdemoapp-1/base.apk"],nativeLibraryDirectories=[/vendor/lib, /system/lib]]
        at dalvik.system.BaseDexClassLoader.findClass(BaseDexClassLoader.java:56)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:511)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:469)
        at com.snowplowanalytics.snowplow.tracker.Tracker$1.run(Tracker.java:497) 
        at android.os.Handler.handleCallback(Handler.java:739) 
        at android.os.Handler.dispatchMessage(Handler.java:95) 
        at android.os.Looper.loop(Looper.java:135) 
        at android.app.ActivityThread.main(ActivityThread.java:5254) 
        at java.lang.reflect.Method.invoke(Native Method) 
        at java.lang.reflect.Method.invoke(Method.java:372) 
[...]

PR #379, opened 2020-05-27
Release/1.4.2

AlexBenny:
This release fix a breaking change introduced in the last version. Also a minor fix to InstallTracker component. 

- Move from Travis CI to GitHub Actions (#375) 
- Fix race condition on tests (#376)
- Fix crash caused by InstallTracker component (#377)
- Version 1.4.1 uses AndroidX instead of Support library (#373)
- Fix broken links to documentation (#368)


.github/workflows/build.yml
```diff
@@ -0,0 +1,65 @@
...
+
+jobs:
+  test:
+    runs-on: macos-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        api-level:
+          - 17
```
paulboocock, line 13: It appears that `reactivecircus/android-emulator-runner@v2` supports down to API Level 15. 

I'd suggest we move to support API Level 16, and test for that here, as this was the first Jelly Bean release. API Level 17 seems a little odd as it was an update to Jelly Bean - unless we need something specifically in 17 onwards? 

Either way, this release should take us back to API Level 14 as it's a minor release then with 1.5.0 we can make the decision to move to 16+. Correct? Unless we're going to say from 1.4.0 we support API Level 16+ only (and we just update the docs to make that clear and also make it clear on 1.4.0 and 1.4.1 that an issue means API Level is 21+).

AlexBenny: Ah right, I actually forgot the reason why I've pushed that to API 17 (so many attempts) I think it was something related to failures not directly related to the code but some weirdness with the testing environment. I can give another try to lower API as meanwhile I've fixed various code issues and maybe they affected the testing at that age.

By the way, I'm not too keen in a perfect match between min tracker's supported version (14) and tested version (hopefully 16 - now 17) because we are doing already a giant step forward testing on three different platforms rather than a single one (24).

I would bump to API 16+ in the tracker 2.0 as it's technically a breaking change.
As long as we are doing minor changes in the tracker I think we can just keep compatibility with API 14 also if not tested. I don't think there are many apps below 16 and we haven't tested our tracker below 16 in our last releases.

AlexBenny: For future reference:

It seems that API 16 fails due to...
```
com.android.ddmlib.InstallException: Unknown failure: Error: Unknown option: -d
usage: pm list packages [-f] [-d] [-e] [-s] [-3] [-i] [-u] [FILTER]
```

Which is probably caused by annoying version incompatibility and other configuration problems. I don't think it's worth to spend more time on that now.
I add a note about the issue in the script - so I don't forget that again!

PR #391, opened 2020-06-16
Renew event ID when event object is reused (close #390)

### LINKED ISSUE #390: Renew event ID when event object is reused
The event ID is set in the event constructor. We want to set the event ID  when the event is tracked. Unless the user has set it programmatically.

AlexBenny:
The event ID is set in the event constructor. We want the event ID  set when the event is tracked. Unless the user has set it programmatically.

### DIFF
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/events/AbstractEvent.java
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/events/AbstractEvent.java
@@ -25,6 +25,7 @@ import com.snowplowanalytics.snowplow.tracker.utils.Util;
 import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.UUID;
 
 /**
  * Base AbstractEvent class which contains common
@@ -36,16 +37,16 @@ import java.util.List;
 public abstract class AbstractEvent implements Event {
 
     protected final List<SelfDescribingJson> customContexts;
-    private final String eventId;
-    protected long deviceCreatedTimestamp;
+    private String eventId;
+    protected Long deviceCreatedTimestamp;
     private Long trueTimestamp;
 
     public static abstract class Builder<T extends Builder<T>> {
 
         private List<SelfDescribingJson> customContexts = new LinkedList<>();
-        private String eventId = Util.getUUIDString();
-        private long deviceCreatedTimestamp = System.currentTimeMillis();
-        private Long trueTimestamp = null;
+        private String eventId;
+        private Long deviceCreatedTimestamp;
+        private Long trueTimestamp;
 
         protected abstract T self();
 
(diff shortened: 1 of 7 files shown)


snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/events/AbstractEvent.java
```diff
@@ -175,11 +182,30 @@ protected AbstractEvent() {
...
+     * Get the timestamp of the event.
+     * @apiNote If the timestamp is not set, it sets one as a side effect.
+     * @deprecated As of release 1.5.0, it will be removed in the version 2.0.0.
+     * @return the event timestamp
      */
     @Override
+    @Deprecated
     public long getDeviceCreatedTimestamp() {
-        return this.deviceCreatedTimestamp;
+        if (deviceCreatedTimestamp == null) {
```
paulboocock, line 193: I'm not sure why we have introduced a side effect in this function and then made a new function to get the unchanged version. Whats the plan for v2?

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/events/AbstractEvent.java
```diff
@@ -175,11 +182,30 @@ protected AbstractEvent() {
...
+            deviceCreatedTimestamp = System.currentTimeMillis();
+        }
+        return deviceCreatedTimestamp;
+    }
+
+    /**
+     * Get the actual timestamp of the event.
+     * @apiNote It doesn't have the side effect of {@link #getDeviceCreatedTimestamp()}.
+     * @deprecated As of release 1.5.0, it will be removed in the version 2.0.0.
```
paulboocock, line 202: It also seems odd to introduce deprecated methods...

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/events/AbstractEvent.java
```diff
@@ -191,11 +217,30 @@ public Long getTrueTimestamp() {
...
+        if (eventId == null) {
+            eventId = Util.getUUIDString();
+        }
+        return eventId;
+    }
+
+    /**
+     * Get the actual event id of the event.
+     * @apiNote It doesn't have the side effect of {@link #getEventId()}.
+     * @deprecated As of release 1.5.0, it will be removed in the version 2.0.0.
```
paulboocock, line 237: Same comments here as above.

AlexBenny:
Thanks @paulboocock , all these comments are related to a design issue.
Normally the event object has a timestamp and an eventId assigned. That can cause duplications if the same object is reused multiple times.
I moved the eventId and timestamp assignment into the tracker rather than the event constructor.
Unfortunately, these are public API so I can't change the contract of the `getEventId` and `getDeviceCreatedTimestamp`. They have to return a non-null value and I have to let the user set them as usual if they want (until v.2.0).

In brief:
- the side effect exists because I can't change the behaviour of the two old get methods
- the two new methods return the "actual" values because Event is an interface and I can access eventId and timestamp only by methods (no properties on interfaces)
- everything is deprecated because the v.2.0 of Event interface will not have any trace of eventID and timestamp, so no need of the old and new methods!

PR #393, opened 2020-06-18
CI script improvements (close #387)

### LINKED ISSUE #387: CI script improvements
Since we moved CI to GitHub Actions we want to improve it progressively:
- Trigger on push rather than (push and pull requests)
- Split the build in two different workflows: regular build and deployment
- Generate API reference documentation

AlexBenny:
- Trigger on push rather than push and pull requests
- Split the build in two different workflows: regular build and deployment
- Generate API reference documentation (using Doxygen)

Output example here: https://snowplow.github.io/snowplow-android-tracker/


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Build
+        run: ./gradlew assembleRelease
+
+      - name: Get version (git tag)
+        id: get_version
+        run: |
+          echo ${GITHUB_REF/refs\/tags\//}
+          echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
+
+      - name: Publish
```
paulboocock, line 32: You can perform all the logic that is in deploy.sh directly in Github Actions if you want to. 
See: https://github.com/snowplow/snowplow-java-tracker/blob/65d3a0cca839ed6dbc111afcaf4bac639fd41c25/.github/workflows/deploy.yml#L43

Totally optional but would remove the `.scripts`.

.github/workflows/doc.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Doxygen
+        uses: mattnotmitt/doxygen-action@v1
+        with:
+          working-directory: '.doxygen'
+          doxyfile-path: './Doxyfile'
+    
+      - name: Deploy to GitHub Pages
+        if: success()
+        uses: crazy-max/ghaction-github-pages@v2
```
paulboocock, line 33: There seem to be some far more popular Actions for deploying pages rather than this one. If this works though then I have no issue with it but thought I'd point it out, as more stars often means better future support.
https://github.com/marketplace/actions/github-pages-action
https://github.com/marketplace/actions/deploy-to-github-pages

PR #395, opened 2020-06-24
Background threads don't report exceptions (close #394)

### LINKED ISSUE #394: Background threads don't report exceptions
When a background threads hit an exception it just fails silently. Nothing is reported to the app or logged anywhere.

Those exception has to be logged in the console and handled by the diagnostic feature.

AlexBenny:
The background threads weren't able to report exceptions correctly. They failed silently and lot of the operations done by the Android tracker happen in background threads.
We need to report those issues in the logs, and eventually with the diagnostic feature, in order to spot in advance anything suspicious in the tracker behaviour.

### DIFF
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/ExecutorTest.java
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/ExecutorTest.java
@@ -1,0 +1,34 @@
+package com.snowplowanalytics.snowplow.tracker;
+
+import android.test.AndroidTestCase;
+
+import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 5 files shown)


snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/ExecutorTest.java
```diff
@@ -0,0 +1,34 @@
...
+            synchronized (expectation) {
+                expectation.notify();
+            }
+        });
+        synchronized (expectation) {
+            expectation.wait(10000);
+        }
+        assertTrue(exceptionRaised.get());
+    }
+}
```
paulboocock, line 34: Missing new line

PR #410, opened 2020-08-19
Emitter refactoring (close #409)

### LINKED ISSUE #409: Emitter refactoring
The current Emitter component covers different roles:
- it bundle events together
- it generates the requests
- it coordinates the persistence and resending of events with the EventStore
- it physically sends the requests to the collector

This monolithic component makes the refactoring and testing quite complex.
The purpose of this task is:
- split Emitter in two parts (Emitter and NetworkConnection)
- use interfaces to improve unit tests
- minor internal refactoring (no changes in the way it works)

AlexBenny:
The Emitter component makes the refactoring and testing quite complex.
The purpose of this PR is:
- split Emitter in two parts (Emitter and NetworkConnection)
- use interfaces to improve unit tests
- minor internal refactoring (no changes in the way it works)

Along with the work on the Emitter I've also touched the EventStore as it was directly connected to the Emitter implementation. Now it's behind an interface of the same name `EventStore` (the implementation is `DefaultEventStore`).

There were a couple of possible issues I've tried to fix:
- The database in the EventStore requires time for the initialisation. During that time an event couldn't be stored. Now I've removed the limitation adding a buffer able to store the events meanwhile the database is starting up.
- When the emission process raise an exception the Emitter could freeze in a `isRunning` state, unable to send other events until it restarts. Now an exception is logged and doesn't freeze the Emitter.


snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/emitter/EmitterEvent.java
```diff
@@ -0,0 +1,24 @@
...
+    public final Payload payload;
+    public final long eventId;
+
+    public EmitterEvent(@NonNull Payload payload, long eventId) {
+        this.payload = payload;
+        this.eventId = eventId;
+    }
+
+    @Override
+    public String toString() {
```
paulboocock, line 18: toString on this object runs the risk of exposing data into logging which might not be desirable. By implementing toString in this way, it becomes much easier for this to happen by mistake - when somewhere in the application the event is used as a string and unexpectadly writes this data in a log to some other system (potentially a third party).
It would be better in my opinion to remove this and directly use the properties (I think this toString is only really used in the tests from what I can see). 
It's a subtle change but one that reduces the footprint for data to leak out on errors/logging.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/storage/DefaultEventStore.java
```diff
@@ -0,0 +1,315 @@
...
+/**
+ * Helper class for storing, getting and removing
+ * events from the SQLite database.
+ *
+ * @apiNote The database has a long startup time.
+ * The events added to the store are placed in a
+ * temporary list meanwhile the database is not
+ * yet ready.
+ */
+public class DefaultEventStore implements EventStore {
```
paulboocock, line 45: Should this be `sqlLiteEventStore`?

paulboocock (approved):
This looks good to me. A very nice refactor and has made some aspects much easier to read/follow. 

Couple of suggestions but once resolved happy for this to be merged.

paulboocock:
One addition note, not really linked to this PR but the addition of `NetworkConnection` made me wonder if we should also upgrade the okHttp3 library? It's pretty out of date, we should either bump to 3.14.9 or even better would be to go to 4.8.1.

AlexBenny:
I've updated OkHttp to 3.12 as the newer versions (even 3.14) cuts off the Android versions below 5.0.
We could move to newer versions in the version 2.0 of the tracker.
It's a mistake I did few months ago and then we needed to revert rapidly to the old version to avoid breaking changes. There I discovered that 3.13 introduces a different min supported Android version!

I've also changed the SQLiteEventStore implementation making `sendLimit` an external parameter as it seems more correct to me and it's implemented in that way in the iOS tracker.

paulboocock:
Good spot on the Android Support! Makes me sad a minor revision of OkHttp3 changed the supported versions.

paulboocock (approved):

Issue #366, opened 2020-04-08
Make session uuid rotation dependant on events being tracked


colmsnowplow:
Some local tests have uncovered slightly unexpected behaviour of the session id incrementation logic. 

While testing the foreground and background timeouts (by setting them to 10s for example), some events came through with a new session id, but previous session id was not attributable to any previous session (see screenshot, timestamp is dvce_created).

I suspect that what's happening here is the foreground and background timeout settings come with a logic of:

"Whenever the timeout expires, rotate uuids"

However in my tests, the timeout expired twice without an event being sent, so the session id in the middle was 'lost'.

If this is the case, we can avoid it by only rotating the uuid at the point of creating an event - so every new session id is attached to an event.

In practice it's fairly unlikely that any reasonable usage for the timeouts would run into this issue.

![image](https://user-images.githubusercontent.com/25744626/78806221-4c58fe00-79ba-11ea-89e5-1ebb79827dc9.png)

AlexBenny:
Some details about the implementation in the same issue on the iOS tracker repo: https://github.com/snowplow/snowplow-objc-tracker/issues/497#issuecomment-694858945

Issue #347, opened 2020-01-30
Session storageMechanism is marked SQLITE but it's not 


AlexBenny:
The session context field `storageMechanism` is always set to `SQLITE` but actually the tracker doesn't use SQLite to persist session data.

See: snowplow/snowplow-objc-tracker#482

PR #413, opened 2020-09-22
Release/1.7.0

AlexBenny:
In this release we refactored the session management which fixes an issue where short session timeout could cause more than one session expiration. The new session management avoids also other possible issues as it adopts a different way to check if the session is expired (more details on #366 thread).

### CHANGELOG

**Bug fixes:**
- Make session uuid rotation dependant on events being tracked (#366)

**Under the hood:**
- Session storageMechanism is marked SQLITE but it's not (#347)


paulboocock (approved):
LGTM. Couple of comments on the comments but other than that, seems good!

PR #427, opened 2021-03-17
Allow multiple instances of the tracker (close #426) 

### LINKED ISSUE #426: Allow multiple instances of the tracker
The upcoming version 2.0 will allow multiple tracker instances in the same app.
It's important to note that from version 2.0 the emitter as a separate component no longer exists.
The tracker is fully build automatically based on the configuration passed. The configuration is attached to the specified namespace. The multiple trackers will behave as completely different trackers with their own configuration and collector.

At the time of writing there isn't a way to send the a same event with a single `track` call to multiple trackers. To send the event to all the trackers the developer has to call `track` on each tracker. The event will be tracker separately by them and they will be received in the pipeline with different `event_id` and `device_timestamp`. This is needed to avoid ambiguities hard to catch in the data modelling phase.

Further iteration on this topic will be done to improve the usability of this feature and cover the most common use cases.

AlexBenny:
This PR divert from the original idea to enforce the tracker as a singleton.
Some third party libraries use the tracker for their tracking. A singleton would prevent the host app to instrument the same tracker along with the tracker in a third party library. Hence we wanted to allow multiple trackers to work in the same app.

In this PR:
- ServiceProvider as manager of the tracker. It instances the classes: tracker, eventStore and emitter (enforcing 1:1:1 constraint).
- Snowplow as entry point to instance the trackers and configure them, or eventually remove them.
- EventStore improved in order to handle multiple sqlite files/databases.
- Improved automatic tracking in order to work with multiple trackers.


snowplow-demo-app/src/main/java/com/snowplowanalytics/snowplowtrackerdemo/Demo.java
```diff
@@ -206,17 +213,19 @@ private void setupAndTrackDemoEvents() {
         Protocol security = _security.getCheckedRadioButtonId() ==
                 _radioHttp.getId() ? Protocol.HTTP : Protocol.HTTPS;
 
-        if (!e.getEmitterStatus()) {
-            e.setEmitterUri(uri);
-            e.setRequestSecurity(security);
-            e.setHttpMethod(method);
+        EmitterController e = Snowplow.getDefaultTracker().getEmitter();
```
paulboocock, line 216: I really quite like this default tracker stuff. I was thinking about implementing diagnostic events in the JS tracker but it's hard without the default tracker concept. I _might_ be stealing this ;)

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/Snowplow.java
```diff
@@ -0,0 +1,140 @@
...
+    @Nullable
+    private ServiceProvider defaultServiceProvider;
+    @NonNull
+    private final Map<String, ServiceProvider> serviceProviderInstances = new HashMap<>();
+
+    // Singleton
+
+    private Snowplow() {
+        if (singleton != null) {
+            throw new RuntimeException("Use `getInstance` method to create this class.");
```
paulboocock, line 38: How is this ever going to be fired? You can't call this constructor without calling `getInstance()`?

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/Snowplow.java
```diff
@@ -0,0 +1,140 @@
...
+        }
+    }
+
+    private static Snowplow getInstance() {
+        if (singleton == null) {
+            synchronized (Snowplow.class) {
+                if (singleton == null) {
+                    singleton = new Snowplow();
```
paulboocock, line 46: I'm not entirely sure why Snowplow is a singleton and not simply full of static members?

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/emitter/storage/EventStoreHelper.java
```diff
@@ -48,7 +55,35 @@
             "dateCreated TIMESTAMP DEFAULT CURRENT_TIMESTAMP)";
 
     // Prevents multiple instances being created and avoids memory leaks.
-    private static EventStoreHelper sInstance;
+    private static HashMap<String, EventStoreHelper> instances = new HashMap<>();
+
+    @NonNull
+    public synchronized static List<String> removeUnsentEventsExceptForNamespaces(@NonNull Context context, @Nullable List<String> allowedNamespaces) {
```
paulboocock, line 61: Whats the use case for passing in `List<String> allowedNamespaces`?

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/InstallTracker.java
```diff
@@ -64,74 +63,27 @@ protected Boolean doInBackground(Context... contexts) {
...
-                .build();
-        try {
-            Tracker.instance().track(event);
-        } catch(IllegalStateException e) {
-            Logger.e(TAG, "Failed to send install event as Tracker is not instanced: %s", e.getMessage());
-        }
-    }
-
-    @Nullable
-    static public SelfDescribingJson getApplicationContext(@NonNull Context context) {
```
paulboocock, line 114: Where did this application context go?

paulboocock (approved):
LGTM

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/InstallTracker.java
paulboocock: I found it. You can ignore this 🤦

snowplow-demo-app/src/main/java/com/snowplowanalytics/snowplowtrackerdemo/Demo.java
AlexBenny: Interesting, about the diagnostic there are some parts I would do differently now. I should probably put some of those ideas/feedbacks in a shared doc to discuss this more widely.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/Snowplow.java
AlexBenny: It can be accessed using reflection. But it just may be overkill.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/Snowplow.java
AlexBenny: It's that because it's the same implementation I did on iOS and I really copied it almost line by line. But you are right... it could be just static methods and props.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/emitter/storage/EventStoreHelper.java
AlexBenny: This method and the way is implemented is something I dislike. The problem is that the app can have zombie events inside EventStores associated to namespace no longer used. In that case the developer at startup can instrument the tracker(s) and delete all the EventStores EXCEPT the ones associated to the "allowedNamespaces", which are the namespaces of the just instrumented trackers.

I think you suggested the right way to go, allowing the zombie events to be associated to emitter configurations and automatically flush them. But it was a too big feature too implement now.

PR #431, opened 2021-03-29
Store sessions info separately per tracker namespace (close #430)

### LINKED ISSUE #430: Store sessions info separately per tracker namespace
Multiple tracker instances needs to have independent session info.
The tracker 1.x used to store some information about the session in the filesystem.
That would cause interferences between trackers in a multi-tracker app.

AlexBenny:
Multiple tracker instances needs to have independent session info.
The tracker 1.x used to store some information about the session in the filesystem.
That would cause interferences between trackers in a multi-tracker app.


paulboocock (approved):

AlexBenny:
@paulboocock I've pushed a new commit with some changes as I noticed some NullPointerException in a flaky test. I've modified Session in order to prepare the file where the session info are stored at the constructor rather than in a separate background task, which was unreliable.

PR #444, opened 2021-04-26
Issue/443 set lifecycle off

AlexBenny:
As planned, I set lifecycle optional.
In order to keep the behaviour similar between Android and iOS, I've also disabled the background/foreground index in case the lifecycle tracking is disabled. In that case all the events will be considered in foreground.

### DIFF
--- a/snowplow-demo-app/build.gradle
+++ b/snowplow-demo-app/build.gradle
@@ -1,7 +1,7 @@
 apply plugin: 'com.android.application'
 
 project.ext {
-    archLifecycleVersion = "1.1.1"
+    archLifecycleVersion = "2.2.0"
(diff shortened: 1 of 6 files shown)


AlexBenny:
@paulboocock I just changed the dependency of lifecycle using "compileOnly" so it can be used to compile and test but it'll be optional for the user in the app. No other changes added.

paulboocock (approved):

paulboocock:
We should make sure the documentation is clear about how to add support for this feature (`implementation "androidx.lifecycle:lifecycle-extensions:$project.archLifecycleVersion"`) for users who want to use Lifecycle Events.

PR #458, opened 2021-06-22
Fix Gdpr context being tracked without enabling it (close #457)

### LINKED ISSUE #457: Fix Gdpr context being tracked without enabling it
**Describe the bug**
Gdpr context is auto tracked even though I haven't configured it

```json
    {
      "schema": "iglu:com.snowplowanalytics.snowplow\/gdpr\/jsonschema\/1-0-0",
      "data": {
        "documentDescription": null,
        "basisForProcessing": "contract",
        "documentId": null,
        "documentVersion": null
      }
    },
```

**To Reproduce**
I'm migrating the existing codebase from 1.7.1 to 2.1.0 I assume it's how Gdpr controller is initialized by default. But if it's actually unexpected to see these Gdpr contexts then I can try to narrow it down in scope and create a minimal viable example.

**Expected behavior**
There is a way to disable Gdpr context as other contexts (platform, session, application, etc.)

**Additional context**
Is it safe/correct for now to disable Gdpr auto context with

```kotlin
  return Snowplow.createTracker(
            context,
            NAMESPACE,
            networkConfiguration,
            trackerConfiguration,
            sessionConfiguration,
            emitterConfiguration,
            subjectConfiguration,
            globalContextsConfiguration
        ).apply {
            gdpr.disable() 
        }
```

paulboocock:
I think this might be a small oversight due to the remote configuration additions in v2.1.0 with the GDPR Context not being part of the remote configuration but adopting a similar pattern when being initialised at tracker creation. The "Update" object is never `null`, it keeps track of only whether it has been updated or has already been configured.

I think this change is pretty safe, looking at the flow current it seems that the only way we need to configure a gdpr context in the builder is if the sourceConfig has been populated yet. I might have missed something though 👀

### DIFF
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/internal/tracker/ConfigurationTest.java
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/internal/tracker/ConfigurationTest.java
@@ -201,6 +201,55 @@ public class ConfigurationTest {
     }
 
     @Test
+    public void withoutGdprConfiguration() throws InterruptedException {
+        MockEventStore eventStore = new MockEventStore();
+        Context context = InstrumentationRegistry.getInstrumentation().getTargetContext();
+        NetworkConfiguration networkConfiguration = new NetworkConfiguration("fake-url", HttpMethod.POST);
+        TrackerConfiguration trackerConfiguration = new TrackerConfiguration("appid")
+                .base64encoding(false);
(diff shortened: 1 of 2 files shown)


AlexBenny (approved):
Thanks @paulboocock , LGTM
I think the problem is that the GDPR configuration is a bit different from the other hence the logic of the others doesn't work well there. Could be iOS is affect by the same problem. I'm going to check it now and release a patch version for both.

Issue #459, opened 2021-06-23
Fix GitHub Actions script with emulator timeout


AlexBenny:
GitHub Actions build workflow doesn't work properly because it's unable to launch the emulator due to timeout error:
https://github.com/ReactiveCircus/android-emulator-runner/issues/160

AlexBenny:
Closing as it's no longer an issue

PR #461, opened 2021-06-24
Release/2.1.1

AlexBenny:
This version has a patch for a bug introduced with the version 2.1. The bug adds the GDPR context to the events even when disabled.
With this fix it will be added only when configured or enabled at runtime.

### CHANGELOG

**Bug fixes:**
- Fix Gdpr context being tracked without enabling it (#457)


paulboocock (approved):

AlexBenny:
@paulboocock Unfortunately tests can't pass because of a bug in the launch of the emulator. I've created an issue for that fix but it's still in progress: #459.
I would suggest to release it anyway. Applying the fix in the master branch as soon as it's ready a fix.

PR #463, opened 2021-07-15
Release/2.2

AlexBenny:
This version reverts a breaking change introduced with the version 2.0.
The application_install event reports the time when the app has been launched for the first time after the installation.
Since the version 2.0 that timestamp is added in the `true_timestamp` field of the event.
Unfortunately, we noticed that it can be cause of issues on data-modelling because the `true_timestamp` is directly applied to the `derived_timestamp` field and it's not adapted to the time of the collector (`collector_timestamp`). The problem is visible when the user device has the device time set on the future, because the `derived_timestamp` wouldn't be corrected and would store the application_install event as happened in the future.

Since this version, we set the `application_install` timestamp in the `device_created_timestamp`. In this way, a wrong time setting in the user device wouldn't affect the correctness of the `derived_timestamp` in the data model.


snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/Tracker.java
```diff
@@ -540,6 +552,14 @@ private Tracker(@NonNull TrackerBuilder builder) {
         this.level = builder.logLevel;
         this.foregroundTimeout = builder.foregroundTimeout;
         this.backgroundTimeout = builder.backgroundTimeout;
+        this.trackerVersionSuffix = builder.trackerVersionSuffix;
+
+        if (trackerVersionSuffix != null) {
+            String suffix = trackerVersionSuffix.replaceAll("[^A-Za-z0-9.\\-]", "");
```
paulboocock, line 558: Any reason we're allowing `\` on Android but not on iOS?

paulboocock (approved):
LGTM. I reran the tests since one seemed to have randomly failed. All the tests also passed locally for me, these flakey tests are definitely something to think about improving in the future 😬

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/Tracker.java
AlexBenny: On iOS I used a different approach without the regex. Here `\` is the escape for the other `\` that is the escape for the dash `-`, which is a special character in that regex. Bit crazy!

paulboocock: You don't need to escape the `\` for the `-`. By adding two `\` you are escaping the `\`. Infact, you don't need to escape `-` at all if you place it at the end of the character class. You can escape it (since it does have a special meaning in a character class) but I actually think it's more conventional to not escape it and place it at the end.
The regex you want is: `[^A-Za-z0-9.-]`
https://regexr.com/620jo

AlexBenny: Azz ... I fix it and release it

Issue #466, opened 2021-08-13
Fix Session UserID not consistent among tracker instances


AlexBenny:
**Describe the bug**
Each tracker instance (identified by different namespaces) has its own different Session UserID.

**Expected behaviour**
The Session UserID should be the same for all the tracker instances in the same app for the same installation.

PR #467, opened 2021-08-13
Release/2.2.1

AlexBenny:
This is a patch version that fixes a bug introduced with the version 2.0 which allows multiple tracker instances in the same app. The Session UserID should be common among the instances but with the previous version each instance of the tracker generated its own Session UserID. With this fix, all the instances can share the same Session UserID.

### CHANGELOG

**Bug fixes:**
- Fix Session UserID not consistent among tracker instances (#466)


paulboocock (approved):
LGTM

PR #472, opened 2021-09-13
Capture campaign attribution data from deep links (close #470)

### LINKED ISSUE #470: Capture campaign attribution data from deep links
The purpose of this enhancement is to create a deep_link_received event that can be manually sent when the app receives a deep_link url. Then, at the first ScreenView, will be attached a deep_link context with the same values of the deep_link_received already sent.

This enhancement is a duplicate of: https://github.com/snowplow/snowplow-objc-tracker/issues/297

AlexBenny:
This PR introduces the deep-linking in the Android tracker.
The process is partially manual.
1. The user tracks a deep-link event passing url and eventually referrer
2. The tracker keeps those info in the internal state
3. At the first ScreenView tracked, the tracker adds a DeepLink entity to the event
(The latter can be disabled from TrackerConfiguration)

**Below the proposed documentation update:**

--
## Deep Linking

In the mobile world the Deep Links are url that links to a specific content directly in a mobile app. Usually they are used to redirect the user from a website or a banner-ad to the installed app instead of another web page. The Deep Links help the user to go straight to the content they are interested of inside the app, similarly to the navigation in the web. 

Usually the Deep Link is received by the mobile operating system and passed to the related app. It's the responsability of the developer to parse the received deep-link's url and show the right content to the user. The way the Deep Link is used is strongly dependant by the way the app routing is implemented. For this reason, our mobile tracker can't automatically track the Deep Link, but we provide an out-of-the-box event that can be used by the developer to manually track it as soon as the Deep Link is received in the app. It will be duty of the tracker to automatically attach the informations of the Deep Link to the first ScreenView tracked.

The [Android platform documentation](https://developer.android.com/training/app-links/deep-linking) provides clear instructions about how to declare the Deep Links that the app is able to handle. Following those instructions, to enable Deep Link support in your app you should declare an *intent filter* in your `AndroidManifest.xml` that sends the Deep Link to the selected Activity as an intent.

In practice, when the activity receives the intent, the developer can track it through the DeepLinkReceived event.
For example (in your Activity):
```
@Override
public void onCreate(Bundle savedInstanceState) {
    ...

    // Extract info from Intent
    Intent intent = getIntent();
    String deepLinkUrl = intent.getData().toString();
    String referrer = null;
    Bundle extras = intent.getExtras();
    if (extras != null) {
        Uri referrerUri = extras.get(Intent.EXTRA_REFERRER);
        if (referrerUri != null) {
            referrer = referrerUri.toString();
        }
    }
    // Create and track the event
    DeepLinkReceived event = new DeepLinkReceived(deepLinkUrl).referrer(referrer);
    tracker.track(event);

    ...
}
```

The tracker keeps memory of the tracked Deep Link event and will attach a Deep Link entity to the first ScreenView tracked in the tracker.
This is helpful during the analysis of the data because it will be clear the relation between the content visualized by the user (ScreenView event) and source (DeepLink entity) that originated that visualization.

This behaviour is enabled by default but it can be disabled from the TrackerConfiguration.

For example:
```
    ...
    .deepLinkContext(false)
    ...
```

### DIFF
--- a/snowplow-demo-app/src/main/java/com/snowplowanalytics/snowplowtrackerdemo/Demo.java
+++ b/snowplow-demo-app/src/main/java/com/snowplowanalytics/snowplowtrackerdemo/Demo.java
@@ -51,6 +51,7 @@ import com.snowplowanalytics.snowplow.configuration.TrackerConfiguration;
 import com.snowplowanalytics.snowplow.controller.EmitterController;
 import com.snowplowanalytics.snowplow.controller.SessionController;
 import com.snowplowanalytics.snowplow.controller.TrackerController;
+import com.snowplowanalytics.snowplow.event.ScreenView;
 import com.snowplowanalytics.snowplow.globalcontexts.GlobalContext;
 import com.snowplowanalytics.snowplow.tracker.DevicePlatform;
 import com.snowplowanalytics.snowplow.tracker.LoggerDelegate;
@@ -298,7 +299,7 @@ public class Demo extends Activity implements LoggerDelegate {
             return;
         }
         TrackerEvents.trackAll(tracker);
-        eventsCreated += 9;
+        eventsCreated += 10;
         final String made = "Made: " + eventsCreated;
         runOnUiThread(() -> _eventsCreated.setText(made));
     }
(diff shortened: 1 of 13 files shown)


AlexBenny:
Link to related PR for schemas: https://github.com/snowplow/iglu-central/issues/1129

adatzer (approved):
Looks great!

Issue #249, opened 2017-08-11
Add tracking of Play Store referer


alexanderdean:
Mixpanel handles it (https://mixpanel.com/help/reference/android#referrer_tracking) by adding a listener and setting it as a super property.

We could add a similar listener and then fire an event - what should the event be called?

htmathias:
Bear in mind, you can pass any string through referrer, but GA tells users to follow UTM parameters with source being a required string, so you could parse this by default.
(https://developers.google.com/analytics/devguides/collection/android/v4/campaigns#google-play-url-builder)

Another point to consider is that we actually use our own custom referrer model rather than UTM, FB and other ad agencies also do this, which is mostly a series of key:value pairs, we'd have to parse that in the SQL modelling or in the listener. Not the end of the world, but we'd maybe have to do our own 'attribution event' anyway with a schema.

alexanderdean:
Thanks @htmathias , that's helpful context...

PR #475, opened 2021-09-22
Add convenient methods for DeepLinkReceived event (close #474)

### LINKED ISSUE #474: Add convenient methods for DeepLinkReceived event
The DeepLinkReceived event has to be filled manually by the developer.
The tracker can't automatically generate the event.
However, we can add some convenient methods to help the developer to create the event starting from some Android SDK objects that have already the info needed to fill the DeepLinkReceived event.

AlexBenny:
I've added a couple of factory methods to generate a DeepLinkReceived event from an Intent or a ReferrerDetails.
The Intent is provided by the Activity that receives the deep link. The developer can create the DeepLinkReceived event just passing the Intent.
Same for the ReferrerDetails. In this case the ReferrerDetails is generated by the Install Referrer Library that can be added as a dependency by the developer. We don't force this dependency and it's optional in the tracker.

Below the suggested documentation (that integrate the one in PR: https://github.com/snowplow/snowplow-android-tracker/pull/472)

---

## Deep Linking

In the mobile world the Deep Links are url that links to a specific content directly in a mobile app. Usually they are used to redirect the user from a website or a banner-ad to the installed app instead of another web page. The Deep Links help the user to go straight to the content they are interested of inside the app, similarly to the navigation in the web. 

Usually the Deep Link is received by the mobile operating system and passed to the related app. It's the responsability of the developer to parse the received deep-link's url and show the right content to the user. The way the Deep Link is used is strongly dependant by the way the app routing is implemented. For this reason, our mobile tracker can't automatically track the Deep Link, but we provide an out-of-the-box event that can be used by the developer to manually track it as soon as the Deep Link is received in the app. It will be duty of the tracker to automatically attach the informations of the Deep Link to the first ScreenView tracked.

The [Android platform documentation](https://developer.android.com/training/app-links/deep-linking) provides clear instructions about how to declare the Deep Links that the app is able to handle. Following those instructions, to enable Deep Link support in your app you should declare an *intent filter* in your `AndroidManifest.xml` that sends the Deep Link to the selected Activity as an intent.

In practice, when the activity receives the Intent, the developer can track it through the DeepLinkReceived event.
For example (in your Activity):
```
@Override
public void onCreate(Bundle savedInstanceState) {
    ...

    // Extract info from Intent
    Intent intent = getIntent();
    String deepLinkUrl = intent.getData().toString();
    String referrer = null;
    Bundle extras = intent.getExtras();
    if (extras != null) {
        Uri referrerUri = extras.get(Intent.EXTRA_REFERRER);
        if (referrerUri != null) {
            referrer = referrerUri.toString();
        }
    }
    // Create and track the event
    DeepLinkReceived event = new DeepLinkReceived(deepLinkUrl).referrer(referrer);
    tracker.track(event);

    ...
}
```

It's also possible to use the factory method `fromIntent` that directly generates the DeepLinkReceived event from an Intent passed.

```
@Override
public void onCreate(Bundle savedInstanceState) {
    ...

    // Generate event directly from an Intent
    Intent intent = getIntent();
    DeepLinkReceived event = DeepLinkReceived.fromIntent(intent);
    // Create and track the event
    if (event != null) {
         tracker.track(event);
    }

    ...
}
```

The deep link can be received by an already installed app. In case the developer wants to receive a deep link of a fresh installed app, they have to use the Install Referrer Library which can provide the original deep link url.

The Install Referrer Library is an Android standard library that can be instrumented as explained [here](https://developer.android.com/google/play/installreferrer/library). The Snowplow Android tracker doesn't force this dependency so the developer has to add the dependency on gradle:

```
dependencies {
    ...
    implementation("com.android.installreferrer:installreferrer:2.2")
}
```

The ReferrerDetails will be accessible as explained [here](https://developer.android.com/google/play/installreferrer/library#install-referrer). In there you can generate the DeepLinkReceived event and track it.

```
@Override
    public void onInstallReferrerSetupFinished(int responseCode) {
        switch (responseCode) {
            case InstallReferrerClient.InstallReferrerResponse.OK:
                ...
                // Generate event directly from an ReferrerDetails
                ReferrerDetails referrerDetails = referrerClient.getInstallReferrer();
                DeepLinkReceived event = DeepLinkReceived.fromReferrerDetails(referrerDetails);
                // Create and track the event
                if (event != null) {
                    tracker.track(event);
                }
                ...
```

Every time a DeepLinkReceived event is tracked, the tracker keeps memory of the tracked Deep Link url and will attach a Deep Link entity to the first ScreenView tracked in the tracker.
This is helpful during the analysis of the data because it will be clear the relation between the content visualized by the user (ScreenView event) and source (DeepLink entity) that originated that visualization.

This behaviour is enabled by default but it can be disabled from the TrackerConfiguration.

For example:
```
    ...
    .deepLinkContext(false)
    ...
```


snowplowcla:

AlexBenny:
A similar issue about the topic (noticed now): https://github.com/snowplow/snowplow-android-tracker/issues/249

paulboocock (approved):
This is a nice set of helper methods 👌

PR #477, opened 2021-10-07
Add memory, battery and storage info to mobile context (closes #476)

### LINKED ISSUE #476: Add memory, battery and storage info to mobile context
We want to extend the range of device information sent as context along with events to automatically include memory, storage and battery properties. Concretely, we identified the following properties to include:

* memory
   * physicalMemory – Total physical system memory in bytes
   * systemAvailableMemory – Available memory on the system in bytes (Android only)
* battery
   * batteryLevel – Remaining battery level as an integer percentage of total battery capacity
   * batteryState – Battery state for the device
* storage
   * availableStorage – Bytes of storage remaining
   * totalStorage – Total size of storage in bytes

The mobile_context schema version 1.0.2 provides properties for all the information (see [issue here](https://github.com/snowplow/iglu-central/issues/1168)).

We decided to add the properties under the same tracker configuration for platform context (platformContext) as done with previous information about device platform (which included network status). We do not expect the new information to incur significant overhead in computation or space and so there is no need to introduce new configuration options.

matus-tomlein:
This pull requests adds new device information to mobile platform context of events as described in #476 

It also refactors the tracker code in order to cache platform context and update it in regular intervals. The architecture is the same as for the iOS tracker.

**Architecture**

I created a class diagram that shows the main classes used in this pull requests and their interaction:
https://docs.google.com/drawings/d/1paaL4_sJktjuM3Q1sU2iVRBho1lqfkR5x4ALCYmbYrc/edit?usp=sharing

There are the following main classes:

1. DeviceInfoMonitor – fetches device information from the OS
2. PlatformContext – manages and updates device information fetched using DeviceInfoMonitor
3. Tracker – adds the context from PlatformContext to tracked events

Tracker holds a reference to a PlatformContext instance that is initialized in its constructor. It queries the PlatformContext for the device context whenever a new event is tracked.

PlatformContext stores a dictionary (Map) with the device context information. The dictionary is populated by calling helpers on an instance of the DeviceInfoMonitor class that is kept in the PlatformContext. The dictionary is updated when the information is requested within specified intervals. A different interval is used for updating information about the device (available memory, storage, battery – updated every 0.1s) than network information (network technology and type – updated every 10s).

DeviceInfoMonitor provides helper functions for fetching the device information. It is also called from the Utils class which used to provide public helpers for getting some of the information – these are now deprecated and call the DeviceInfoMonitor.

**Testing**

Tested the demo app on a few devices on Browserstack. Also should be covered by unit tests for the most part.

### DIFF
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/utils/DeviceInfoMonitor.java
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/utils/DeviceInfoMonitor.java
@@ -37,43 +37,26 @@ import java.lang.reflect.Method;
 
 public class DeviceInfoMonitor {
 
-    /**
-     * @return the OS Type
-     */
     @NonNull
     public String getOsType() {
         return "android";
     }
 
-    /**
-     * @return the OS Version
-     */
     @NonNull
     public String getOsVersion() {
         return android.os.Build.VERSION.RELEASE;
     }
 
-
-    /**
-     * @return the device model
-     */
     @NonNull
     public String getDeviceModel() {
         return android.os.Build.MODEL;
     }
 
-    /**
-     * @return the device vendor
-     */
     @NonNull
     public String getDeviceVendor() {
         return android.os.Build.MANUFACTURER;
     }
 
-    /**
-     * @param context the android context
-     * @return a carrier name or null
-     */
     @Nullable
     public String getCarrier(@NonNull Context context) {
         TelephonyManager telephonyManager =
@@ -114,12 +97,6 @@ public class DeviceInfoMonitor {
         }
     }
 
-    /**
-     * Returns the network type that the device is connected to
-     *
-     * @param networkInfo The NetworkInformation object
-     * @return the type of the network
-     */
     @NonNull
     public String getNetworkType(@Nullable NetworkInfo networkInfo) {
         String networkType = "offline";
@@ -136,12 +113,6 @@ public class DeviceInfoMonitor {
         return networkType;
     }
 
-    /**
-     * Returns the network technology
-     *
-     * @param networkInfo The NetworkInformation object
-     * @return the technology of the network
-     */
     @Nullable
     public String getNetworkTechnology(@Nullable NetworkInfo networkInfo) {
         String networkTech = null;
@@ -154,12 +125,6 @@ public class DeviceInfoMonitor {
         return networkTech;
     }
 
-    /**
-     * Returns an instance that represents the current network connection
-     *
-     * @param context the android context
-     * @return the representation of the current network connection or null
-     */
     @Nullable
     public NetworkInfo getNetworkInfo(@NonNull Context context) {
         ConnectivityManager cm = (ConnectivityManager)
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/PlatformContext.java
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/PlatformContext.java
@@ -65,11 +65,6 @@ public class PlatformContext {
         this(100, 10 * 1000, new DeviceInfoMonitor(), context);
     }
 
-    /**
-     * Returns the Mobile Context
-     *
-     * @return the mobile context
-     */
     @Nullable
     public SelfDescribingJson getMobileContext() {
         updateEphemeralDictsIfNecessary();


snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/Tracker.java
```diff
@@ -732,7 +736,7 @@ private void addBasicContextsToContexts(@NonNull List<SelfDescribingJson> contex
         }
 
         if (mobileContext) {
-            contexts.add(Util.getMobileContext(this.context));
+            contexts.add(platformContext.getMobileContext());
```
AlexBenny, line 739: 🎉

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/utils/DeviceInfoMonitor.java
```diff
@@ -0,0 +1,319 @@
...
+
+        int batteryLevel = (int) (level * 100 / (float)scale);
+
+        return new Pair<>(batteryState, batteryLevel);
+    }
+
+    /**
+     * Checked in the user data directory.
+     * @return Currently available storage on disk in bytes
+     */
```
AlexBenny, line 241: This comments are great, but don't feel that you have to add api comments to all the methods.
IMO the javadoc on the public API is already enough. We can get rid of comments in the internal classes even if their methods are public methods. Actually, I would completely exclude them from the Java doc as those classes shouldn't be used by the final users. Though, I think comments are very valuable in cases where the behaviour is hard to guess. In this case, your code is very well written and easy to read!
Btw, I'm just sharing my point of view, so no need to remove the comment in this PR now.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/utils/Util.java
```diff
@@ -305,87 +308,53 @@ static public SelfDescribingJson getApplicationContext(@NonNull Context context)
...
-            return null;
-        }
-    }
-
-    /**
+     * @deprecated Moved to internal API.
      * @return the OS Type
      */
     @NonNull
     public static String getOsType() {
```
AlexBenny, line 315: I'm not sure that the `@deprecated` tag in the Javadoc is enough to inform the compiler.
I guess we should add also the `@Deprecated` annotation in the signature in order to inform the compiler and raise the warning at build time.

Also, for the doc, we inform that we moved it.
What if we just say that it will be removed in the v3?
We don't want they use the internal classes, so in case they complain we can decide to leave these methods in the public API for v3 too.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/utils/Util.java
```diff
@@ -504,64 +432,12 @@ public static boolean mapHasKeys(@NonNull Map<String, Object> map, @NonNull Stri
      *              the key
      * @param map the map to insert the pair into
      */
-    public static void addToMap(@NonNull String key, @NonNull Object value, @NonNull Map<String, Object> map) {
+    public static void addToMap(@Nullable String key, @Nullable Object value, @NonNull Map<String, Object> map) {
```
AlexBenny, line 435: 👍

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/internal/tracker/PlatformContextTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+import android.test.AndroidTestCase;
+
+import com.snowplowanalytics.snowplow.internal.constants.Parameters;
+import com.snowplowanalytics.snowplow.internal.constants.TrackerConstants;
+import com.snowplowanalytics.snowplow.payload.SelfDescribingJson;
+
+import java.util.Map;
+
+public class PlatformContextTest extends AndroidTestCase {
```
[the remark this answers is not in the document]
matus-tomlein: Changed it, thanks!

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/utils/DeviceInfoMonitor.java
matus-tomlein: I agree that some of the methods are pretty self-explanatory and the comments don't add much. I removed some such comments for the internal classes. I'm not sure whether removing all comments for internal methods is a good idea, sometimes they add useful information for us. It seems to be possible to [exclude packages from Javadoc configuration](https://maven.apache.org/plugins/maven-javadoc-plugin/examples/exclude-package-names.html), so we can also try that.

PR #490, opened 2021-12-13
Do not allow multiple state machines with the same ID (close #489)

### LINKED ISSUE #489: Do not allow multiple state machines with the same ID
The problem occurs when multiple state machines are registered with the same identifier after each other. In this case, both state machines will be active and stay in the internal `schemaRegistry` in StateManager but only the last one will be referenced in the `identifierToStateMachine` and `stateMachineToIdentifier` dictionaries.

In practice, the problem occurs when you call `trackerController.setScreenContext(true)` multiple times. In this case, each event will contain multiple instances of the `screen` context entity.

matus-tomlein:
Addresses issue #489 

The change automatically removes previous state machine (if any) with the same identifier when adding a new one. This prevents multiple of the same context being added to events in case the same state machine is registered multiple times.

### DIFF
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/internal/tracker/StateManagerTest.java
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/internal/tracker/StateManagerTest.java
@@ -36,7 +36,7 @@ public class StateManagerTest {
     @Test
     public void testStateManager() {
         StateManager stateManager = new StateManager();
-        stateManager.addStateMachine(new MockStateMachine(), "identifier");
+        stateManager.addOrReplaceStateMachine(new MockStateMachine(), "identifier");
 
         SelfDescribing eventInc = new SelfDescribing("inc", new HashMap() {{ put("value", 1); }});
         SelfDescribing eventDec = new SelfDescribing("dec", new HashMap() {{ put("value", 2); }});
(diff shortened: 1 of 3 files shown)


snowplowcla:

AlexBenny (commented):
Thanks for the fix, great job!

I think we need to consider this issue: when the tracker is reconfigured (passing a new configuration or with remote configuration) the state machine is replaced. Unfortunately the replacing operation [resets the internal state](https://github.com/snowplow/snowplow-android-tracker/blob/74cefe5dc39652f9ac41304328dc964f2c7cc1b0/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/StateManager.java#L42)  for that state machine causing the loss of the state.

I think [here](https://github.com/snowplow/snowplow-android-tracker/blob/74cefe5dc39652f9ac41304328dc964f2c7cc1b0/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/StateManager.java#L27) we could check if the previous state machine and the new state machine have the same type. In that case we don't reset the state machine and we don't loose the internal state.

wdyt?

matus-tomlein:
@AlexBenny That's a really good point, thanks! I updated the StateManager to check for the type of state machine being replaced and do nothing in case it is being replaced with one of the same type. Please take another look at it when you have the chance to see if it's okay.

AlexBenny (approved):
LGTM! 👍 🚢

PR #493, opened 2021-12-23
Fix updating IDFA during app runtime (close #492)

### LINKED ISSUE #492: Fix updating IDFA during app runtime
The IDFA is retrieved when tracker is initialized but never updated. This is a problem in case the app starts while the tracker isn't allowed to access the IDFA but is later given permission from the user – the tracker will cache nil as the IDFA while the app is running.

Problem introduced in v3.

matus-tomlein:
Addresses issue #492 by updating the Android IDFA while it is null or empty. Once a value for it is found, it will no longer be updated.


snowplowcla:

AlexBenny (approved):
LGTM :shipit: 

My unique doubt (not a blocker now) is about the _android context_. PlatformContext keeps the _android context_ but we need to pay attention because IIRC the _android context_ might change during the app life. If PlatformContext retain the old _android context_ that could be source of problems. Probably not a problem in this case but it's just a reminder for future better investigation.

PR #496, opened 2022-01-05
Fix multiple ScreenContexts on ScreenView events (close #495)

### LINKED ISSUE #495: Fix multiple ScreenContexts on ScreenView events
**Describe the bug**
The ScreenContext should be just one for each event. The version 3.0 introduced a bug where the ScreenView event can be potentially tracked with multiple ScreenContexts of the same type.

**To Reproduce**
1. Track a ScreenView
2. Track another ScreenView
3. Check number of ScreenContexts in the last ScreenView event (they should be 2)

**Expected behavior**
Just one ScreenContext for each ScreenView event.

AlexBenny:
As described in the #495 issue the ScreenView events could be tracked with multiple ScreenContexts.
This is caused by the erroneous management of the list of state machines.
During the computation of the entities to add to the event, the tracker gets a list of state machines for the specific event schema, then it gets a list of state machines that can be used with each event (filter: "*"). Unfortunately, the latter was added directly to the first list causing duplication. More ScreenView tracked caused even more duplications.

I've fixed it making a clone of the list rather than using that directly.
The iOS tracker is not affected because it correctly makes a clone.

### DIFF
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/internal/tracker/StateManagerTest.java
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/internal/tracker/StateManagerTest.java
@@ -129,6 +129,7 @@ public class StateManagerTest {
         entities = (String) payload.getMap().get("co");
         assertNotNull(entities);
         assertTrue(entities.contains("screen1"));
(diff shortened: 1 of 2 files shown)


snowplowcla:

matus-tomlein (approved):
Really good catch, I would never think of that! It looks good to me, I don't have any comments.

PR #501, opened 2022-02-07
Add session callback (close #500)

### LINKED ISSUE #500: Add session callback
We removed the session callbacks in the tracker v2 with the plan to reintroduce them back with an enhanced solution.
There are various cases where it's useful to be aware of the current session id or when a session is updated by the tracker.
This new solution should be able to provide a session state in the callback with all the session details tracked by the tracker.

AlexBenny:
Issue: #500 

The proposed session callback makes use of the SessionState which contains all the information normally available in the session context.
The current implementation is tailor-made on the specific problem of getting session information but in a further enhancement it could be generalised as callback that returns information about any internal tracker state currently kept private (e.g: ScreenState, LifecycleState, etc.).

Proposed documentation:

--

## Session Context

Client session tracking is activated by default but it can be disabled through the TrackerConfiguration as explained above. When enabled the tracker appends a [client_session](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-1) context to each event it sends and it maintains this session information as long as the application is installed on the device.

Sessions correspond to tracked user activity. A session expires when no tracking events have occurred for the amount of time defined in a timeout (by default 30 minutes). The session timeout check is executed for each event tracked. If the gap between two consecutive events is longer than the timeout the session is renewed. There are two timeouts since a session can timeout in the foreground (while the app is visible) or in the background (when the app has been suspended, but not closed).

### Session callback

(Available from v3.1)

The tracker allows the configuration of a callback to inform the app everytime a new session is created (in correspondence of a session timeout check). 
This can be configured in the `SessionConfiguration` and it provides the `SessionState` where can be accessed all the info already tracked in the `SessionContext`.

Below an example of where the session callback is used to print out the values of session every time a new session is generated by the tracker:

```
...
SessionConfiguration sessionConfig = new SessionConfiguration(
                new TimeMeasure(6, TimeUnit.SECONDS),
                new TimeMeasure(30, TimeUnit.SECONDS)
        )
                .onSessionUpdate(state -> log("Session: " + state.getSessionId()));
...
Snowplow.createTracker(getApplicationContext(),
                namespace,
                networkConfiguration,
                sessionConfiguration
);
```

### DIFF
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/configuration/SessionConfiguration.java
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/configuration/SessionConfiguration.java
@@ -1,8 +1,11 @@
 package com.snowplowanalytics.snowplow.configuration;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.core.util.Consumer;
 
 import com.snowplowanalytics.snowplow.internal.session.SessionConfigurationInterface;
+import com.snowplowanalytics.snowplow.tracker.SessionState;
 import com.snowplowanalytics.snowplow.util.TimeMeasure;
 
 import org.json.JSONObject;
@@ -38,6 +41,12 @@ public class SessionConfiguration implements SessionConfigurationInterface, Conf
      */
     @NonNull
     public TimeMeasure backgroundTimeout;
+
+    /**
+     * The callback called everytime the session is updated.
+     */
+    @Nullable
+    public Consumer<SessionState> onSessionUpdate;
 
     // Constructors
 
@@ -85,6 +94,34 @@ public class SessionConfiguration implements SessionConfigurationInterface, Conf
     @Override
     public void setBackgroundTimeout(@NonNull TimeMeasure backgroundTimeout) {
         this.backgroundTimeout = backgroundTimeout;
+    }
+
+    /**
+     * @see #onSessionUpdate
+     */
+    @Nullable
+    @Override
+    public Consumer<SessionState> getOnSessionUpdate() {
+        return onSessionUpdate;
+    }
+
+    /**
+     * @see #onSessionUpdate
+     */
+    @Override
+    public void setOnSessionUpdate(@Nullable Consumer<SessionState> onSessionUpdate) {
+        this.onSessionUpdate = onSessionUpdate;
+    }
+
+    // Builders
+
+    /**
+     * @see #onSessionUpdate
+     */
+    @NonNull
+    public SessionConfiguration onSessionUpdate(@Nullable Consumer<SessionState> onSessionUpdate) {
+        this.onSessionUpdate = onSessionUpdate;
+        return this;
     }
 
     // Copyable
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/session/Session.java
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/session/Session.java
@@ -155,8 +155,11 @@ public class Session {
 
     private synchronized static String retrieveUserId(Context context, SessionState state) {
         String userId = state != null ? state.getUserId() : Util.getUUIDString();
-        // Get or Set the Session UserID
-        //$ Specify why this mess with the userId
+        // With v2 we designed a new identifier: Installation ID.
+        // It should be created by the tracker at first execution and it should be constant until the app deletion.
+        // It behaves as the SessionUserID but it should be available even if the session context is disabled.
+        // At the moment we store it separately from the session in order to fully implement it in one of the
+        // future versions.
         SharedPreferences generalPref = context.getSharedPreferences(TrackerConstants.SNOWPLOW_GENERAL_VARS, Context.MODE_PRIVATE);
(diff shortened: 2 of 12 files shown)


snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/session/Session.java
```diff
@@ -292,47 +312,75 @@ int getForegroundIndex() {
...
+            sessionMap.put(Parameters.SESSION_PREVIOUS_ID, null);
+            sessionMap.put(Parameters.SESSION_STORAGE, "LOCAL_STORAGE");
+            return sessionMap;
+        } finally {
+            StrictMode.setThreadPolicy(oldPolicy);
+        }
+    }
+
+    @Nullable
+    private Map<String, Object> getSessionMapFromLegacyTrackerV3(@NonNull Context context, @NonNull String sessionVarsName) {
```
matus-tomlein, line 365: This is quite off-topic but just wanted to suggest that, in the future, we can look into deserialization of maps or JSONs into POJO (maybe using the Jackson library). Probably it's just about preference, but I find this dictionary parsing a bit hard to follow and check. I think defining a simple class for what is expected and then mapping the Map to it would be more explicit. I'm always happier to work with POJO than maps.

matus-tomlein (approved):
Looks great!

Is there any chance that we could add a test for the session callback being fired? Something like the `testSessionConfigurationCallbackAfterNewSession` in the Obj-C tracker. Don't know if there are any technical complications to implement that here, just asking.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/configuration/SessionConfiguration.java
```diff
@@ -87,6 +96,34 @@ public void setBackgroundTimeout(@NonNull TimeMeasure backgroundTimeout) {
...
+
+    /**
+     * @see #onSessionUpdate
+     */
+    @NonNull
+    public SessionConfiguration onSessionUpdate(@Nullable Consumer<SessionState> onSessionUpdate) {
+        this.onSessionUpdate = onSessionUpdate;
+        return this;
+    }
+
```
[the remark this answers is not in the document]
AlexBenny: ah damn! good spot!

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/session/Session.java
```diff
@@ -123,45 +128,48 @@ public Session(long foregroundTimeout, long backgroundTimeout, @NonNull TimeUnit
...
+    }
 
-        // Get or Set the Session UserID
+    private synchronized static String retrieveUserId(Context context, SessionState state) {
+        String userId = state != null ? state.getUserId() : Util.getUUIDString();
+        // With v2 we designed a new identifier: Installation ID.
+        // It should be created by the tracker at first execution and it should be constant until the app deletion.
+        // It behaves as the SessionUserID but it should be available even if the session context is disabled.
+        // At the moment we store it separately from the session in order to fully implement it in one of the
+        // future versions.
```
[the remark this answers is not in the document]
AlexBenny: I guess it's a bit confusing. Let me reword it

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/session/Session.java
AlexBenny: I agree with the general point and some of the previous refactoring went exactly in this direction.
I have just a couple of concerns: 
- I wouldn't add other dependencies in the library. Jackson is definitely cool and widely used but we don't own the final app where the tracker will be instrumented and any dependency we add can be a source of issues in the app and for us with the maintenance.
- SessionState could be considered the POJO but I didn't want to add the serialization code there. On iOS tracker I created SPDataPersitence and that seems a bit cleaner than this.

My doubt is, If we avoid Jackson lib, would the code we write for serialization/deserialization be much different than this?
I think much of this complexity would fade away when (if) we transform the Session manager in a FSM like I did for ScreenContext and DeepLinks. Unfortunately it was off-topic and a too long shot for this task.

matus-tomlein (approved):
Looks great! Nice tests and thanks for clarifying the Installation ID behavior, makes sense to me now.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/session/Session.java
matus-tomlein: We should discuss this further somewhere when we have a chance. I agree that introducing additional dependencies is not ideal. Many modern languages (Swift, Kotlin) already have support for this built in. But if there is no easy way currently to do it in Java, it might be better to hold off on the idea rather than adding a dependency.

I do believe that there is a difference between writing code to parse a dictionary (or JSON) and mapping a dictionary onto an interface (POJO class). If we avoid writing the parsing code and only define what we expect and let the framework map it, it can reduce the potential for bugs, make the code more explicit and easier to read. But I definitely agree that it is not the only way to write good code, we sure can write easy to follow code with the current approach as well.

PR #504, opened 2022-02-17
Check session for background event using foreground timeout (close #499)

### LINKED ISSUE #499: Check session for background event using foreground timeout
Same as https://github.com/snowplow/snowplow-objc-tracker/issues/667

AlexBenny:
Not much to add to what is already described in the related issue (#499).

For Android this fix is much more complex because the event processing and the session management is different.
This is another part that should be fixed with the implementation of the session manager above the already implemented FSM system used for ScreenView management.

The solution I've adopted here is to move the session context attribution in an early stage of the event processing in order to avoid any race condition due to parallel processing of the events. Doing that we are able to reliably manage the sequence of the events. In particular we can check the session for `application_foreground` and `application_background` with the correct session timeout. It wouldn't be possible in a pure asynchronous approach that doesn't take care of the order the events enter in the tracker.

Suggested documentation:

---

## Session Context

Client session tracking is activated by default but it can be disabled through the TrackerConfiguration as explained above. When enabled the tracker appends a client_session context to each event it sends and it maintains this session information as long as the application is installed on the device.

Sessions correspond to tracked user activity. A session expires when no tracking events have occurred for the amount of time defined in a timeout (by default 30 minutes). The session timeout check is executed for each event tracked. If the gap between two consecutive events is longer than the timeout the session is renewed. There are two timeouts since a session can timeout in the foreground (while the app is visible) or in the background (when the app has been suspended, but not closed).

--added part--

The lifecycle events (`application_foreground` and `application_background` events) have a role in the session expiration. The lifecycle events can be enabled in the [TrackerConfiguration](https://docs.snowplowanalytics.com/docs/collecting-data/collecting-from-own-applications/mobile-trackers/mobile-trackers-v3-0/introduction/#TrackerConfiguration) enabling `lifecycleAutotracking` (Note: on Android it requires `androidx.lifecycle:lifecycle-extensions`). Once enabled they will be fired automatically when the app moves from foreground state to background state and vice versa.

When the app moves from foreground to background the `application_background` event is fired. If session tracking is enabled, the session context will be attached to the event checking the session expiration using the foreground timeout.
When the app moves from background to foreground the `application_foreground` event is fired. If session tracking is enabled, the session context will be attached to the event checking the session expiration using the background timeout.

For instance, with this configuration:

```
SessionConfiguration(
    TimeMeasure(360L, TimeUnit.SECONDS),
    TimeMeasure(15L, TimeUnit.SECONDS)
)
```

the session would expire if the app is backgrounded for more than 15 seconds, like in this example:

```
time: 0s - screen_view event - foreground timeout session check - session 1
time: 3s - application_background event - foreground timeout session check (3 < 360) - session 1
time: 30s - application_foreground event - background timeout session check (30 > 15) - session 2
```

In the above example the `application_foreground` event triggers a new session because the time spent on background (without tracked events) is bigger than the background timeout for the session.

### DIFF
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/session/Session.java
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/session/Session.java
@@ -56,15 +56,15 @@ public class Session {
 
     // Session Variables
     private String userId;
-    private int backgroundIndex = 0;
-    private int foregroundIndex = 0;
+    private volatile int backgroundIndex = 0;
+    private volatile int foregroundIndex = 0;
     private SessionState state = null;
 
     // Variables to control Session Updates
     private final AtomicBoolean isBackground = new AtomicBoolean(false);
     private long lastSessionCheck;
     private final AtomicBoolean isNewSession = new AtomicBoolean(true);
-    private boolean isSessionCheckerEnabled;
+    private volatile boolean isSessionCheckerEnabled;
     private long foregroundTimeout;
     private long backgroundTimeout;
 
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/session/Session.java
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/session/Session.java
@@ -258,22 +258,17 @@ public class Session {
     }
 
     /**
-     * Updates the session with information about lifecycle tracking.
+     * Updates the session timeout and the indexes.
      * Note: Internal use only.
-     * @param isForeground whether or not the application moved to foreground.
-     * @return foreground or background index. Returns -1 if the lifecycle state is not changed.
+     * @param isBackground whether or not the application moved to background.
      */
-    public synchronized int updateLifecycleNotification(boolean isForeground) {
-        boolean toBackground = !isForeground;
-        // if the new lifecycle state confirms the session state, there isn't any lifecycle transition
-        if (isBackground.get() == toBackground) {
-            return -1;
+    public void setBackground(boolean isBackground) {
+        if (!this.isBackground.compareAndSet(!isBackground, isBackground)) {
+            return;
         }
-        Logger.d(TAG, "Application is in the background: %s", toBackground);
-        isBackground.set(toBackground);
 
-        if (!toBackground) {
-            Logger.d(TAG, "Application moved to foreground, starting session checking...");
+        if (!isBackground) {
+            Logger.d(TAG, "Application moved to foreground");
             this.executeEventCallback(this.foregroundTransitionCallback);
             try {
                 setIsSuspended(false);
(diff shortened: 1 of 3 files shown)


snowplowcla:

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/session/Session.java
```diff
@@ -258,35 +258,28 @@ public void startNewSession() {
...
-     * @param isForeground whether or not the application moved to foreground.
-     * @return foreground or background index. Returns -1 if the lifecycle state is not changed.
+     * @param isBackground whether or not the application moved to background.
      */
-    public synchronized int updateLifecycleNotification(boolean isForeground) {
-        boolean toBackground = !isForeground;
-        // if the new lifecycle state confirms the session state, there isn't any lifecycle transition
-        if (isBackground.get() == toBackground) {
-            return -1;
+    public void setBackground(boolean isBackground) {
```
matus-tomlein, line 265: I like that you renamed `updateLifecycleNotification` function to `setBackground`, was quite confusing to me before! But not sure why isn't it synchronized anymore? Is it no longer necessary?

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/Tracker.java
```diff
@@ -698,6 +701,30 @@ private void workaroundForCampaignAttributionEnrichment(@NonNull Payload payload
...
+     is executed in a separate thread. At the moment, the Session management is performed by the legacy
+     solution that doesn't make use of the tracker state snapshot (taken as soon as the event is tracked)
+     needed to keep a coherent state between events tracked in sequence but processed in parallel.
+     Due to this parallel processing the session context could be incoherent with the event sequence.
+     This limit is a serious problem when we process lifecycle events because the `application_background`
+     event should be checked with the session foreground timeout and the `application_foreground` event
+     should be checked with the session background timeout. Without a coherent session management it's
+     impossible to grant this behaviour causing serious issues on the calculation of the session expiring.
+     */
+    private void workaroundForIncoherentSessionContext(@NonNull TrackerEvent event) {
```
matus-tomlein, line 715: This is a bit of a shame that we need to introduce the special handling, but I understand the problem. Would be great to support tracker state snapshot in the future.

matus-tomlein (approved):
Looks good and seems to work well! 👍

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/session/Session.java
AlexBenny: I was in doubt but now this method is simplified. The unique operations of writing are on booleans and integer. `isBackground` is an AtomicBoolean, the others are just set and get (no compare) so shouldn't need any synchronization. However, I think it would be good to make them `volatile` in order to pass the info among threads.

PR #506, opened 2022-02-21
Add pause and resume to EmitterController (close #505)

### LINKED ISSUE #505: Add pause and resume to EmitterController
This would support low bandwidth situations when emitting events should be paused.

matus-tomlein:
* adds `pause` and `resume` functions to `EmitterController` that call `pauseEmit` and `resumeEmit` on the Emitter


snowplowcla:

AlexBenny (approved):
Looks great, thanks!

There is a missing piece to add. The `isPaused` in `EmitterConfigurationUpdate` is needed to resume the proper state of the Emitter in case a new configuration is submitted. So, if the emitter was paused and the configuration is refreshed, the emitter is paused again, otherwise a refreshed configuration would override the pause action of the user.

For the tracker this is managed in the `ServiceProvider` here: https://github.com/snowplow/snowplow-android-tracker/blob/bc9f21e6d9ce0f652662143ae79a43e13857a3e7/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/ServiceProvider.java#L422-L424

We should do the same for the Emitter.

matus-tomlein:
@AlexBenny Thanks, I was wondering about where it is used, but then forgot about the issue. I made the change but is there any way that I could test it?

AlexBenny:
> I made the change but is there any way that I could test it?

Short answer: There aren't unit tests for this (configuration refresh in general)

Long answer: Actually, It's something that would be useful as it would make any changes in the configuration much more robust. An easy way to test it would be configure the tracker using a MockNetworkConnection (like you did) and  pause the emitter. Then, refresh the configuration of the tracker with the same namespace and check if the emitter is running or it's still paused.
However, the best would be to have a configurable SPServiceProvider where we can replace the various components with simple mocks created by us, where we can check if the methods `pause` and `resume` have been called.

matus-tomlein:
Added a test for updating the configuration in `ServiceProvider` – it actually uncovered a bug, had to move to the call to pause the Emitter earlier because some waiting events could be tracked while the configuration was updated.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/ServiceProvider.java
```diff
@@ -388,6 +388,9 @@ private Emitter makeEmitter() {
     @NonNull
     private Tracker makeTracker() {
         Emitter emitter = getEmitter();
+        if (emitterConfigurationUpdate.isPaused) {
+            emitter.pauseEmit();
+        }
```
AlexBenny, line 393: Maybe this part can be moved in `makeEmitter` as it's pertinent to the emitter. I guess the behaviour wouldn't change as it updates the emitter as soon as the emitter is instanced.

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/ServiceProviderTest.java
```diff
@@ -0,0 +1,61 @@
...
+import com.snowplowanalytics.snowplow.event.Structured;
+import com.snowplowanalytics.snowplow.internal.emitter.EmitterConfigurationUpdate;
+import com.snowplowanalytics.snowplow.internal.tracker.ServiceProvider;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ServiceProviderTest extends AndroidTestCase {
+
+    public void testUpdatingConfigurationRetainsPausedEmitter() throws InterruptedException {
```
AlexBenny, line 32: Thanks, this was very much needed!

AlexBenny (approved):
👍 👍

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/ServiceProvider.java
matus-tomlein: Good point, it belongs to `makeEmitter`, moved it there.

PR #509, opened 2022-03-04
Enable updating collector endpoint while Emitter is running (close #508)

### LINKED ISSUE #508: Enable updating collector endpoint while Emitter is running
**Describe the bug**
Updating the tracker endpoint doesn't work.

**To Reproduce**
I tried 2 things:

1- Having an instance of the `TrackerController`, which is initialized at startup:
```
trackerController.network?.endpoint = NEW_ENDPOINT
```

2- Using the default tracker controller:
```
Snowplow.getDefaultTracker()?.network?.endpoint = NEW_ENDPOINT
```

In both cases I still see requests being made to the same old endpoint.

I'm using Snowplow version 3.0.3.

**Expected behavior**
The endpoint is updated and events are sent to `NEW_ENDPOINT`.

matus-tomlein:
This PR addresses issue #508 which reported that updating the network endpoint doesn't have any effect. I found that there are two possible reasons for this:

1. the Emitter is currently running, or
2. it uses a custom network connection.

I don't think that anything should happen in the second case. It sounds reasonable that custom network connections are not updated by setting the endpoint in tracker controller (let me know if you think otherwise).

But the first case is less intuitive and can be a problem for several reasons:

- the Tracker calls `flush()` on Emitter right after initialisation which means that the Emitter starts running and it is not possible to update the endpoint right after the tracker is initialised, and
- Emitters can be running for quite a long time. Even when all events are sent, `attemptEmit()` function goes to sleep a number of times checking for new events.

So I think it's reasonable to enable updating the network configuration while the Emitter is running. This PR enables that and removes the check from setters for network configuration in Emitter. I didn't want to have the network connection change during emitting and so I made the `attemptEmit` function accept a `NetworkConnection` to be used while it is running. The network connection is only refreshed in the `attemptEmit` after the sleep while waiting for new events.

I wrapped the `networkConnection` instance variable in an atomic reference. I don't think it's extra necessary in this case, probably it's overly cautious.

### DIFF
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/EmitterTest.java
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/EmitterTest.java
@@ -206,9 +206,9 @@ public class EmitterTest extends AndroidTestCase {
 
         assertTrue(emitter.getEmitterStatus());
         emitter.setHttpMethod(POST);
-        assertEquals("https://com.acme/i", emitter.getEmitterUri());
-        emitter.setEmitterUri("com/foo");
-        assertEquals("https://com.acme/i", emitter.getEmitterUri());
+        assertEquals("https://com.acme/com.snowplowanalytics.snowplow/tp2", emitter.getEmitterUri());
+        emitter.setEmitterUri("com.foo");
+        assertEquals("https://com.foo/com.snowplowanalytics.snowplow/tp2", emitter.getEmitterUri());
         emitter.setBufferOption(DefaultGroup);
         assertEquals(HeavyGroup, emitter.getBufferOption());
 
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/EmitterTest.java
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/EmitterTest.java
@@ -377,6 +377,18 @@ public class EmitterTest extends AndroidTestCase {
         emitter.flush();
     }
 
+    public void testUpdatesNetworkConnectionWhileRunning() throws InterruptedException {
+        Emitter emitter = new Emitter(getContext(), "com.acme", new Emitter.EmitterBuilder()
+                .eventStore(new MockEventStore()));
+
+        emitter.flush();
+        Thread.sleep(100);
+        assertTrue(emitter.getEmitterStatus()); // is running
+        emitter.setEmitterUri("new.uri"); // update while running
+        assertTrue(emitter.getEmitterStatus()); // is running
+        assertTrue(emitter.getEmitterUri().contains("new.uri"));
+    }
+
     // Emitter Builder
 
     public Emitter getEmitter(NetworkConnection networkConnection, BufferOption option) {
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/emitter/Emitter.java
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/emitter/Emitter.java
@@ -475,7 +475,7 @@ public class Emitter {
         emptyCount = 0;
 
         List<EmitterEvent> events = eventStore.getEmittableEvents(sendLimit);
-        List<Request> requests = buildRequests(events);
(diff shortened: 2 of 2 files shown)


snowplowcla:

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/EmitterTest.java
```diff
@@ -206,9 +206,9 @@ public void testUpdatingEmitterSettings() throws InterruptedException {
 
         assertTrue(emitter.getEmitterStatus());
         emitter.setHttpMethod(POST);
-        assertEquals("https://com.acme/i", emitter.getEmitterUri());
-        emitter.setEmitterUri("com/foo");
-        assertEquals("https://com.acme/i", emitter.getEmitterUri());
+        assertEquals("https://com.acme/com.snowplowanalytics.snowplow/tp2", emitter.getEmitterUri());
+        emitter.setEmitterUri("com.foo");
+        assertEquals("https://com.foo/com.snowplowanalytics.snowplow/tp2", emitter.getEmitterUri());
```
matus-tomlein, line 211: I had to change the behavior of this test as it was testing that the network config doesn't change while emitter is running.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/emitter/Emitter.java
```diff
@@ -468,7 +469,7 @@ private void attemptEmit() {
             } catch (InterruptedException e) {
                 Logger.e(TAG, "Emitter thread sleep interrupted: " + e.toString());
             }
-            attemptEmit();
+            attemptEmit(getNetworkConnection()); // at this point we update network connection since it might be outdated after sleep
```
AlexBenny, line 472: Good idea this comment 👍

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/emitter/Emitter.java
```diff
@@ -512,7 +513,7 @@ private void attemptEmit() {
             Logger.e(TAG, "Emitter loop stopping: failures.");
             isRunning.compareAndSet(true, false);
         } else {
-            attemptEmit();
+            attemptEmit(networkConnection);
```
AlexBenny, line 516: Wondering if `getNetworkConnection` is suitable here too 🤔

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/emitter/Emitter.java
```diff
@@ -530,7 +531,7 @@ private void attemptEmit() {
     protected List<Request> buildRequests(@NonNull List<EmitterEvent> events) {
         List<Request> requests = new ArrayList<>();
         String sendingTime = Util.getTimestamp();
-        HttpMethod httpMethod = networkConnection.getHttpMethod();
+        HttpMethod httpMethod = getNetworkConnection().getHttpMethod();
```
AlexBenny, line 534: Maybe we could pass `httpMethod` as argument of `buildRequests`?
I'm thinking about the case where `buildRequests` is called inside `attemptEmit` but meanwhile the networkConnection changes. Apparently `buildRequests` and `isOversize` just need the httpMethod.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/emitter/Emitter.java
```diff
@@ -599,7 +600,7 @@ private boolean isOversize(@NonNull Payload payload) {
      * @return weather the payload bundle exceeds the maximum size allowed.
      */
     private boolean isOversize(@NonNull Payload payload, @NonNull List<Payload> previousPaylods) {
-        long byteLimit = networkConnection.getHttpMethod() == GET ? byteLimitGet : byteLimitPost;
+        long byteLimit = getNetworkConnection().getHttpMethod() == GET ? byteLimitGet : byteLimitPost;
```
AlexBenny, line 603: Same of `buildRequests`

AlexBenny (approved):
Looks great! 👍 

I've just added a couple of comments about `buildRequests`.
Also a quick heads up if you haven't noticed it: this PR doesn't merge on the release branch.

PR #511, opened 2022-04-22
Set log level without trackerDiagnostic (close #441)

### LINKED ISSUE #441: Set log level without trackerDiagnostic
Hello. Our team use SnowPlow analytics for Android application.
In last release (1.7.1) logs are not working, we can not see them in LogCat in Android Studio. We used SnowPlow version 1.3.3 before and on that version logs were visible.
What's the reason of this behavior? Maybe there are any special new flags for logging now?

mscwilson:
For issue #441.

During Tracker construction, setting of the logging level had mistakenly been put inside a conditional. The effect was that, since logging is OFF by default, no logs at all could been seen unless `trackerDiagnostic` had also been specified in the TrackerBuilder options.

This change allows logs to be set properly, and also adds tests for logging.

### DIFF
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/Tracker.java
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/Tracker.java
@@ -515,8 +515,9 @@ public class Tracker {
             if (level == LogLevel.OFF) {
                 level = LogLevel.ERROR;
             }
-            Logger.updateLogLevel(level);
         }
+
+        Logger.updateLogLevel(level);
 
         // When session context is enabled
         if (this.sessionContext) {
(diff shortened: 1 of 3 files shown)


snowplowcla:

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/Tracker.java
```diff
@@ -515,9 +515,10 @@ public Tracker(@NonNull TrackerBuilder builder) {
             if (level == LogLevel.OFF) {
```
AlexBenny, line 515: I think we could also flat these two nested if with an `&&`

AlexBenny (approved):
👍 Great! Thanks for adding the test!

PR #515, opened 2022-05-03
 Attach different entity in demo global context (close #514)

### LINKED ISSUE #514: Attach different entity in demo global context
Events generated by the built-in demo are getting two identical Application entities attached. There should only be one Application entity per event.

mscwilson:
For issue #514.
Two copies of the Application entity were being added to events generated by the demo. This turned out to be because applicationContext was turned on, but the same entity (and data) was being used to demonstrate global context.

This change uses a different schema for the global contexts example.


snowplowcla:

AlexBenny (approved):
Thanks for fixing this! 👍 Now it's much better.

Issue #512, opened 2022-04-29
Attach LifecycleEntity before the first Background event


mscwilson:
When LifecycleAutotracking is configured, the tracker automatically generates Background events when the app is hidden, and Foreground events when it comes back to the foreground.

Also, a LifecycleEntity is automatically added to all events describing whether the app is visible or not. Or, it's supposed to be. Currently, LifecycleEntities are only added once the first Background event has been generated, as that's when the Lifecycle state is set.

LifecycleEntities should be attached to all events: before the first Background event, the app is in the foreground/visible, and events should be labelled as such.

PR #516, opened 2022-05-23
Release/3.1.2

mscwilson:
# Version 3.1.2

This patch version fixes a couple of problems relating to context entities, and a logging bug.

Logging can be set as part of the tracker configuration, but is off by default in the Android tracker. Unfortunately, a logic error meant that log levels would only be set if `trackerDiagnostic` mode was also configured. Logging now works as expected.

The first entity problem relates to the Lifecycle autotracking feature. When configured, `Background` and `Foreground` events are fired as the user navigates away from, and back to, the app. Also, a `LifecycleEntity` is attached to all events, marking if they occurred when the app was visible or not. Or rather, that's what now correctly happens with this release. Previously, lifecycle entities were mistakenly not added to events until after a Background event had fired.

The minor second problem was in the built-in demo app. Two copies of the application entity were being added to the generated events. This has been fixed by using a different schema for the global context example.

## CHANGELOG

**Bug fixes:**
* Set log level without trackerDiagnostic (#441)
* Attach LifecycleEntity before the first Background event (#512)
* Attach different entity in demo global context (#514)


snowplowcla:

mscwilson:
Internal PR for [release notes](https://github.com/snowplow/engineering-announcements/pull/34).

AlexBenny (approved):
Perfect 👍 🚢

PR #521, opened 2022-06-13
 Add eventIndex and firstEventTimestamp properties to client sessions (close #517) 

### LINKED ISSUE #517: Add eventIndex and firstEventTimestamp properties to client sessions
Add the new properties of the 1-0-2 version of the `client_session` schema.

* eventIndex – Optional index of the current event in the session. Signifies the order of events in which they were tracked.
* firstEventTimestamp – Optional date-time timestamp of when the first event in the session was tracked.

These properties are optional to avoid a breaking schema change.

mscwilson:
For issue #517.

Adds the new properties of the 1-0-2 version of the client_session schema.

- EventIndex – Index of the current event in the session, starting from 1.
- FirstEventTimestamp – Datetime timestamp of when the first event in the session was tracked.

This change also adds a test for the session context entity.

### DIFF
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/SessionState.java
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/SessionState.java
@@ -9,6 +9,7 @@ import com.snowplowanalytics.snowplow.internal.tracker.State;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
 
 public class SessionState implements State {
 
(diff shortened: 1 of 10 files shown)


snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/tracker/SessionState.java
```diff
@@ -13,41 +13,46 @@ public class SessionState implements State {
...
     @NonNull
-    private final String storage;
-    @NonNull
     private final String userId;
+    @NonNull
+    private final String storage;
 
     @NonNull
-    private Map<String, Object> sessionContext;
+    private final Map<String, Object> sessionContext;
```
AlexBenny, line 29: 👍 Nice add! - btw, maybe you are already aware of this but, just in case... making it final will not make the map immutable. It will make immutable only the reference to the map.

mscwilson: Yeah that makes sense. IntelliJ was hassling me to make it final.

AlexBenny (approved):
👍 👍 LGTM!

PR #523, opened 2022-06-23
Add eventIndex and firstEventTimestamp properties to client sessions (close #517)

### LINKED ISSUE #517: Add eventIndex and firstEventTimestamp properties to client sessions

matus-tomlein:
There was a small issue in the `Util.getDateTimeFromTimestamp()` method implementation – it didn't consider the timezone and thus could return local time although the timestamp was acting like it was UTC (it ended with `Z`). So I made a small change and added this on line 70 in Util.java:

```java
        dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
```

### DIFF
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/internal/utils/UtilTest.java
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/internal/utils/UtilTest.java
@@ -13,6 +13,11 @@ public class UtilTest extends AndroidTestCase {
         assertEquals(13, Util.getTimestamp().length());
     }
 
+    public void testGetDateTimeFromTimestamp() {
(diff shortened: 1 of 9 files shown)


AlexBenny (approved):
LGTM 👍 - I've checked just the line you added as I suppose all the other changes come from the previous implementation of the same issue. Let me know if there is more to review.

matus-tomlein:
Thanks @AlexBenny! It's only that one line, sorry for the difficult to read changes.

PR #525, opened 2022-06-24
Add a customizable set of failure HTTP status codes for which collector requests should not be retried (close #518)

### LINKED ISSUE #518: Add a customizable set of failure HTTP status codes for which collector requests should not be retried
Issue description copied from the equivalent [iOS tracker issue](https://github.com/snowplow/snowplow-objc-tracker/issues/684):

Instead of retrying collector requests for all non-2xx status codes, disable the retry functionality for the following HTTP status codes by default:

Code | Description | Rationale
-- | -- | --
400 | Bad Request | Server rejected to process the request due to an issue with it, trying again won't help.
401 | Unauthorized (RFC 7235) | Request is lacking correct authorization or refusing to process some IP address – intentional denial of request. Similar to 403 but specific to authorization issues.
403 | Forbidden | Server is refusing to take action, request should not be repeated.
410 | Gone | Indicates an intentional denial of the request as the resource was removed, request should not be retried and potentially the tracker should stop emitting events.
422 | Unprocessable Entity (WebDAV; RFC 4918) | Server was unable to process events in the request, should not be repeated without modification of the request.

All other non-2xx status codes should be retried on.

The list of HTTP status codes for which to retry or not-retry should be user customizable (e.g., the user should be able to force retries on 403 status code).

matus-tomlein:
This PR addresses issue #518 .

It adds a set of 5 HTTP status codes (400, 401, 403, 410, and 422) for which sending events should not be retried and the events should be dropped. This is in line with the strategy for handling HTTP status codes that we [outlined here](https://github.com/snowplow-incubator/data-value-resources/blob/main/30%20Tracker%20Architecture/emitter_architecture/handling_http_status_codes.md).

It also makes the retry behaviour configurable. It adds the `customRetryForStatusCodes` property to `EmitterConfiguration`. This is a dictionary mapping status codes (int) to booleans stating whether to retry or not.

The only change in the documentation was to add a short note about the property – [here is the PR in data-value-resources](https://github.com/snowplow-incubator/data-value-resources/pull/112).

### DIFF
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/RequestResultTest.java
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/RequestResultTest.java
@@ -1,16 +1,25 @@
 package com.snowplowanalytics.snowplow.tracker;
 
-import android.test.AndroidTestCase;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertEquals;
+
+import androidx.test.ext.junit.runners.AndroidJUnit4;
 
 import com.snowplowanalytics.snowplow.network.RequestResult;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
 
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 
-public class RequestResultTest extends AndroidTestCase {
+@RunWith(AndroidJUnit4.class)
+public class RequestResultTest {
 
+    @Test
     public void testSuccessfulRequest() {
         RequestResult result = new RequestResult(200, false, Arrays.asList(100L));
         assertTrue(result.isSuccessful());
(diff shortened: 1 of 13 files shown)


snowplowcla:

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/RequestResultTest.java
```diff
@@ -0,0 +1,54 @@
...
+import android.test.AndroidTestCase;
+
+import com.snowplowanalytics.snowplow.network.RequestResult;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+public class RequestResultTest extends AndroidTestCase {
```
AlexBenny, line 12: Same comment of the other PR - AndroidTestCase is deprecated.
We should migrate all the tests to the new way, soon or later.

AlexBenny (approved):
👍 👍

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/RequestResultTest.java
matus-tomlein: Changed!

Issue #526, opened 2022-06-27
Update Android target SDK version to 31


matus-tomlein:
Update:

* compileSdkVersion from 30 to 31
* buildToolsVersion from "29.0.3" to "30.0.2"
* targetSdkVersion from 30 to 31

Issue #520, opened 2022-06-10
Bump okhttp3 to 4.10.0, play-services-analytics to 18.0.1, and gson to 2.9.0


mscwilson:
Some security vulnerabilities have been linked to the current version 4.9.1

Issue #519, opened 2022-06-10
Add support for storing cookies in NetworkConnection


mscwilson:
This will allow the `network_userid` to persist across requests.

PR #529, opened 2022-06-28
Release/3.2.0

matus-tomlein:
The tracker now makes use of the  [`client_session` schema version 1-0-2](http://iglucentral.com/schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-2), which adds two new properties to the session context entity – `eventIndex` (index of the event in the session) and `firstEventTimestamp` (timestamp of the first event in the session). The event index property is useful for ordering events as they were tracked.

The release also changes the behaviour for 5 HTTP status codes (400, 401, 403, 410, and 422) that, when received in response from the Collector, will now cause the tracker not to retry sending events (i.e., drop the events). Requests with all other 3xx, 4xx, and 5xx status codes are retried. The set of status codes for which events should be retried or not is customizable in `EmitterConfiguration`. The following shows the configuration in the Android tracker:

```java
EmitterConfiguration emitterConfiguration = new EmitterConfiguration();
Map<Integer, Boolean> retryRules = new HashMap<>();
retryRules.put(403, true); // retry for the 403 HTTP status code
emitterConfiguration.setCustomRetryForStatusCodes(retryRules);
```

Finally, the release adds support for storing cookies set by the Collector. This means that the `network_userid` property set by the Collector is now kept consistent among events tracked by the same tracker. Cookies are stored in shared preferences with an option to provide a custom cookie jar.

## CHANGELOG

**Features:**

* Add eventIndex and firstEventTimestamp properties to client sessions (#517)
* Add a customizable set of failure HTTP status codes for which collector requests should not be retried (#518)
* Add support for storing cookies in NetworkConnection (#519)

**Under the hood:**

* Update Android target SDK version to 31 (#526)
* Bump okhttp3 to 4.10.0, play-services-analytics to 18.0.1, and gson to 2.9.0 (#520)


AlexBenny (approved):
👍 🚢

PR #534, opened 2022-07-26
Do not resume remote configuration from cache in case it has different endpoint (close #533)

### LINKED ISSUE #533: Do not resume remote configuration from cache in case it has different endpoint
When new remote configuration is setup, the tracker first restores any previous configuration from cache while the remote configuration is downloaded. This is a good feature in case similar remote configuration as previously used is being retrieved.

However, some users may switch between remote configurations from different endpoints (e.g., when the user switches country settings in app). In this case, it is not desirable that the previous cached configuration is used. Instead, we should use the default configuration.

The cache should still be used when setting up a new remote configuration with the same endpoint.

[Here is the same issue on the iOS tracker](https://github.com/snowplow/snowplow-objc-tracker/issues/688).

matus-tomlein:
Addresses issue #533 which goal was to support switching between remote configurations with different endpoints. This was not previously possible because the tracker cached previous configurations and didn't consider the endpoint when using the cache. Thus, if one switched to a new remote configuration with a new endpoint, a cache for the previous one could still be used by the tracker.

The approach that I chose is to add a hash of the remote config endpoint URL into the cache file name. So instead of `.../snowplow-cache/remoteConfig.data`, the cache files are now called `.../snowplow-cache/remoteConfig-12498219.data` where the number (12498219) is the hashcode of the URL. I make use of the String hashCode function which is not the most reliable but I think it's sufficient for this purpose.

### DIFF
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/internal/remoteconfiguration/RemoteConfigurationTest.java
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/internal/remoteconfiguration/RemoteConfigurationTest.java
@@ -119,11 +119,12 @@ public class RemoteConfigurationTest {
         expected.configurationVersion = 12;
         expected.configurationBundle = Lists.newArrayList(bundle);
 
-        ConfigurationCache cache = new ConfigurationCache();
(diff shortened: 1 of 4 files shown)


snowplowcla:

AlexBenny (approved):
👍 👍

PR #541, opened 2022-08-22
Add anonymous tracking features (close #540)

### LINKED ISSUE #540: Add anonymous tracking features
Anonymous tracking is a feature currently implemented in the JavaScript tracker that enables anonymising various user and session identifiers. It affects the following user and session identifiers:

1. Client-side user identifiers: `domain_userid` (equivalent to `client_session.userId` in session context). On mobile, there are also the IDFA identifiers in the `mobile_context` platform context entity.
2. Client-side session identifiers: `domain_sessionid` (equivalent to `client_session.sessionId` in session context)
3. Server-side user identifiers: `network_userid` and `user_ipaddress`

There are several levels to the anonymisation depending on which of the three categories are affected:

## 1. Full client-side anonymisation

In this case, both the client-side user identifiers as well as the client-side session identifiers are anonymised.

### Current options in mobile trackers

One could disable tracking both the `client_session` and `mobile_context` context entities:

```
trackerConfig.sessionContext = false
trackerConfig.platformContext = false
```

### TODO in mobile trackers

Disabling the platform context altogether may be a bit wasteful as there is other information (e.g., memory usage, carrier, OS) that are not user or session identifiers (although they could be used for fingerprinting). We could provide an option to only disable the IDFA identifiers in the platform context:

```
trackerConfig.userAnonymisation = true
```

This could set them to null (they are not required in the platform context schema).

## 2. Client-side anonymisation with session tracking

In this case, client-side user identifiers should be disabled but session identifiers should still be tracked.

### Current options in mobile trackers

This is currently not possible in the mobile trackers.

### TODO in mobile trackers

We could provide an option to enable session context while disabling all user identifiers (both the ones in `client_session` and `mobile_context` context entities):

```
trackerConfig.sessionContext = true
trackerConfig.userAnonymisation = true
```

The `userId` in `client_session` is a required property, so this could always set it to `00000000-0000-0000-0000-000000000000`.

## 3. Server-side anonymisation

This disables setting the server-side cookies with `network_userid` and also anonymizes the user IP address.

### Current options in mobile trackers

This is currently not possible in the mobile trackers.

### TODO in mobile trackers

We could add an option to the tracker configuration:

```
trackerConfig.serverAnonymisation = true
```

This would set the `SP-Anonymous: *` HTTP header for requests going to the Collector. The collector will take care of not setting the request headers and anonymizing the IP address.

matus-tomlein:
This PR addresses issue #540 and adds anonymous tracking features to the tracker.

Two properties were added in configuration:

* `TrackerConfiguration.userAnonymisation`
  * Disables user identifiers in session context (sets `userId` to `00000000-0000-0000-0000-000000000000`) and platform context (removes `androidIdfa`).
  * It is implemented by masking the identifiers at the time when the entities are added to events. So the `userId` is actually stored in session state but removed when adding to events (JS tracker does a similar thing).
* `EmitterConfiguraiton.serverAnonymisation`
  * Sets the `SP-Anonymous: *` header in requests to Collector (at the time when the requests are sent, not when the events are tracked). This results in anonymising the `network_userid` and `user_ipaddress` properties.
  * Decided to put it in the Emitter config and not Tracker config as it is handled by the Emitter. But if it makes more sense to put it in Tracker config, we can move it there.

# Documentation

(Available from v4)

Anonymous tracking enables anonymising various user and session identifiers. It affects the following user and session identifiers:

1. Client-side user identifiers: the `userId` in Session context entity and the IDFA identifiers (`openIdfa`, `appleIdfa`, and `androidIdfa`) in the Platform context entity.
2. Client-side session identifiers: `sessionId` in Session context.
3. Server-side user identifiers: `network_userid` and `user_ipaddress` event properties.

There are several levels to the anonymisation depending on which of the three categories are affected:

## 1. Full client-side anonymisation

In this case, both the client-side user identifiers as well as the client-side session identifiers are anonymised. To enable it, disable session context and enable user anonymisation:

```java
    .sessionContext(false)
    .userAnonymisation(true);
```

## 2. Client-side anonymisation with session tracking

This setting disables client-side user identifiers are but tracks session information. In practice, this means that events track the Session context entity but the `userId` property is a null UUID (`00000000-0000-0000-0000-000000000000`). In case Platform context is enabled, the IDFA identifiers will not be present.

```java
    .sessionContext(true)
    .userAnonymisation(true);
```

## 3. Server-side anonymisation

Server-side anonymisation affects user identifiers set server-side. In particular, these are the `network_userid` property set in server-side cookie and the user IP address. You can anonymise the properties using the `serverAnonymisation` flag in `EmitterConfiguration`:

```java
EmitterConfiguraiton config = new EmitterConfiguraiton()
    .serverAnonymisation(true);
```

### DIFF
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/internal/tracker/ConfigurationTest.java
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/internal/tracker/ConfigurationTest.java
@@ -1,5 +1,7 @@
 package com.snowplowanalytics.snowplow.internal.tracker;
 
+import static com.snowplowanalytics.snowplow.network.HttpMethod.GET;
+
 import android.content.Context;
 import android.content.SharedPreferences;
 
@@ -21,16 +23,22 @@ import com.snowplowanalytics.snowplow.event.Timing;
 import com.snowplowanalytics.snowplow.globalcontexts.GlobalContext;
 import com.snowplowanalytics.snowplow.emitter.EmitterEvent;
 import com.snowplowanalytics.snowplow.internal.constants.TrackerConstants;
+import com.snowplowanalytics.snowplow.internal.emitter.Emitter;
 import com.snowplowanalytics.snowplow.internal.emitter.Executor;
 import com.snowplowanalytics.snowplow.network.HttpMethod;
 import com.snowplowanalytics.snowplow.network.Protocol;
+import com.snowplowanalytics.snowplow.network.Request;
 import com.snowplowanalytics.snowplow.payload.Payload;
 import com.snowplowanalytics.snowplow.payload.SelfDescribingJson;
 import com.snowplowanalytics.snowplow.tracker.BuildConfig;
 import com.snowplowanalytics.snowplow.tracker.MockEventStore;
+import com.snowplowanalytics.snowplow.tracker.MockNetworkConnection;
 import com.snowplowanalytics.snowplow.util.Basis;
 import com.snowplowanalytics.snowplow.util.TimeMeasure;
 
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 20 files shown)


snowplowcla:

AlexBenny (approved):
LGTM! I don't see issues with this implementation. I've only added a comment on something with very little importance.

PR #545, opened 2022-08-30
Return tracked event ID from the track method (close #544)

### LINKED ISSUE #544: Return tracked event ID from the track method
The `trackerController.track(event)` method should return the tracked event ID in the `UUID` format to surface the event reference for users. This would be in line with the API provided by the Snowplow Java and C++ tracker.

matus-tomlein:
Addresses issue #544 and makes the track method return the event ID as a `UUID`. The change is quite small as the ID was already present in the method, it was only necessary to return it.

### DIFF
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/Tracker.java
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/Tracker.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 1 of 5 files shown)


snowplowcla:

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/Tracker.java
```diff
@@ -632,6 +634,7 @@ public void track(final @NonNull Event event) {
             this.emitter.add(payload);
             event.endProcessing(this);
         });
+        return trackerEvent.eventId;
```
AlexBenny, line 637: 👍 I think in the future we could remove this asynchronous part about event processing. It shouldn't be so expensive in my opinion. Even because on iOS it's not async. I've always thought to make the iOS processing async like on Android, but now I'm more keen to go in the opposite direction, making Android event processing synchrnous until the payload reaches the EventStore.
Just a note for future brainstorming.

AlexBenny (approved):
LGTM! :shipit:

PR #551, opened 2022-10-03
Fix serialization of session timeouts in remote configuration cache (close #547)

### LINKED ISSUE #547: Fix serialization of session timeouts in remote configuration cache
The `SessionConfiguration` stores foreground and background timeouts using `snowplow.util.TimeMeasure` objects. This causes problems when a configuration needs to be serialized to be written to cache on disk as the class is not serializable.

We should make the `snowplow.util.TimeMeasure` class serializable or use a different type to store the timeouts in session configuration.

matus-tomlein:
Issue #547 

This is a tiny PR that just makes the `TimeMeasure` class serializable as it is used in `SessionConfiguration` which needs to be serialized when storing in cache. Otherwise, storing it in cache throws an `IOException`.


snowplowcla:

AlexBenny (approved):
👍 👍

PR #552, opened 2022-10-19
Log an error when recreating tracker after being removed (close #548)

### LINKED ISSUE #548: Log an error when recreating tracker after being removed
Currently, when one removes a tracker using `Snowplow.removeTracker(tracker)` (this closes the tracker) and then calls for instance `tracker.getNamespace()`, the tracker is reinitialized and starts tracking again (e.g., subscribes to automatic screen view tracking). This is unexpected behaviour as the user didn't request the tracker to be reinitialized.

The problem is in the implementation of the `getTracker()` method in `ServiceProvider` (and `TrackerController` in consequence) which calls `makeTracker()` in case the tracker reference is null to initialize it. We should not make a new tracker if it was previously removed.

matus-tomlein:
Issue #548 

The problem that we are dealing with here is that users got confusing behaviour after having code that did something similar to this:

```java
TrackerController tracker = Snowplow.createTracker(...);
Snowplow.removeTracker(tracker);
System.out.println("Tracker removed: " + tracker.getNamespace());
```

In this case the tracker was removed on the second line but it was reinitialized on the third line, which is not obvious based on the snippet. Reinitializing the tracker also means that various autotracking features may be reactivated and unexpected events tracked.

I would like to prevent this behaviour and just return NULL from the `getNamespace()` and other functions from the `TrackerController` but I realized that this would be a breaking change. We would need to change the signature of the `TrackerController` functions.

So in order to prevent making the breaking change, this PR:

1. Logs an error when a tracker that was previously removed is accessed. This should make it easier for users to debug the weird behaviour.
2. Renames some of the internal functions to make clear that they are not just getters but can also change the objects (e.g., `getTracker()` -> `getOrMakeTracker()`).

### DIFF
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/ServiceProvider.java
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/ServiceProvider.java
@@ -23,7 +23,6 @@ import com.snowplowanalytics.snowplow.internal.emitter.NetworkConfigurationInter
 import com.snowplowanalytics.snowplow.internal.emitter.NetworkConfigurationUpdate;
 import com.snowplowanalytics.snowplow.internal.emitter.NetworkControllerImpl;
 import com.snowplowanalytics.snowplow.internal.gdpr.Gdpr;
-import com.snowplowanalytics.snowplow.internal.gdpr.GdprConfigurationInterface;
 import com.snowplowanalytics.snowplow.internal.gdpr.GdprConfigurationUpdate;
 import com.snowplowanalytics.snowplow.internal.gdpr.GdprControllerImpl;
 import com.snowplowanalytics.snowplow.internal.globalcontexts.GlobalContextsControllerImpl;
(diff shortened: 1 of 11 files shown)


snowplowcla:

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/ServiceProvider.java
```diff
@@ -233,79 +232,95 @@ private void initializeConfigurationUpdates() {
     // Getters
 
     @NonNull
-    public Subject getSubject() {
+    @Override
+    public Subject getOrMakeSubject() {
```
AlexBenny, line 236: 👍 👍

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/tracker/ServiceProvider.java
```diff
@@ -233,79 +232,95 @@ private void initializeConfigurationUpdates() {
...
         }
         return emitter;
     }
 
     @NonNull
-    public Tracker getTracker() {
+    @Override
+    public Boolean isTrackerInitialized() {
+        return tracker != null;
+    }
```
AlexBenny, line 256: Same thought I already added on iOS Tracker.
Rethinking to this, maybe we could call this method "isAlive" or "isActive" or in negative form "isInvalidated", related to the fact that it's the ServiceProvider that should be dead but it's still used.
Just a thought btw.

AlexBenny (approved):
LGTM! 👍 👍  Same thought already reported for iOS Tracker.

PR #554, opened 2022-11-02
Add DeepLink entity referrer and url to atomic properties in ScreenView events (close #553)

### LINKED ISSUE #553: Add DeepLink entity referrer and url to atomic properties in ScreenView events
The [Referer parser enrichment](https://docs.snowplow.io/docs/enriching-your-data/available-enrichments/referrer-parser-enrichment/) doesn't parse the DeepLink context entity that is attached to ScreenView events. In order to workaround this, the issue aims to set the `url` and `referrer` properties based on the DeepLink entity also in the atomic event properties.

matus-tomlein:
Issue #553 

This PR extends the workaround in the tracker which sets the page referrer and page URL atomic properties in events – previously it only set them for deep link events based on their properties. Now, it also sets them for screen view events based on the deep link context entities.

I had to move the workaround to a different place. Previously it was called in `addSelfDescribingPropertiesToPayload()` in Tracker but at that point the events do not contain context entities assigned using state machines. Since the deep link entity is added in a state machine, I moved the call to after state machine context entities are added to events.

## Documentation: Tracking Events > Deep Link

The tracked `DeepLinkReceived` event and each subsequent `ScreenView` event also have the URL and referrer information added in the `page_url` and `page_referrer` atomic properties in the events. This makes them compatible with data models built for events tracked on the Web.

### DIFF
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/event/DeepLinkReceivedTest.java
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/event/DeepLinkReceivedTest.java
@@ -27,6 +27,8 @@ import org.junit.runner.RunWith;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.UUID;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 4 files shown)


snowplowcla:

AlexBenny (approved):
LGTM! 
I think we should probably add something in the documentation about this new feature.
I would add the doc changes in the PR description so we don't forget it at the release.

PR #555, opened 2022-11-03
Anonymise previous session ID and user identifiers in subject when user anonymisation is enabled (close #549)

### LINKED ISSUE #549: Anonymise previous session ID and user identifiers in subject when user anonymisation is enabled
In addition to anonymising the `userId`, we should also anonymise the `previousSessionId` attribute in the `client_session` context entity when user anonymisation is enabled. Otherwise, one can recreate the user journey by traversing through the previous session references.

The `previousSessionId` property is optional, so we can just set it to null.

Also anonymize Subject properties including the user id, domain and network user id, and IP address.

matus-tomlein:
Issue #549 

Anonymises additional properties when `TrackerConfiguration.userAnonymisation` is enabled. In particular, the following additional properties are anonymised:

* Session context entity
   * `previousSessionId` – otherwise one could recreate the whole user journey from the chain of sessions and basically recreate a user id
* Subject
   * `userId` – also anonymised in the JavaScript tracker
   * `domainUserId` – also anonymised in the JavaScript tracker
   * `networkUserId` – anonymised when server anonymisation is enabled, I think it makes sense to also anonymise if it is set client-side
   * `ipAddress` – anonymised when server anonymisation is enabled, I think it makes sense to also anonymise if it is set client-side

## Documentation

Anonymous tracking is a tracker feature that enables anonymising various user and session identifiers to support user privacy in case consent for tracking the identifiers is not given.

On mobile, the following user and session identifiers can be anonymised:

* Client-side user identifiers:
   * `userId` in the [Session](http://iglucentral.com/schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-2) context entity
   * IDFA identifiers (`appleIdfa` and `appleIdfv` for iOS, `androidIdfa` for Android) in the [Platform](http://iglucentral.com/schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-2) context entity
   * `userId`, `domainUserId`, `networkUserId`, `ipAddress` if they are set in `Subject`
* Client-side session identifiers: `sessionId` and `previousSessionId` in Session entity.
* Server-side user identifiers: `network_userid` and `user_ipaddress` event properties.

### DIFF
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/SubjectTest.java
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/SubjectTest.java
@@ -41,7 +41,7 @@ public class SubjectTest {
 
     @Test
     public void testGetSubjectStandardPairs() throws Exception {
-        Subject subject = getSubject();
+        Subject subject = createSubject();
         Map<String, String> standardPairs = subject.getSubject(false);
 
         assertTrue(standardPairs.containsKey("tz"));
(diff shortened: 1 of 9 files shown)


snowplowcla:

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/SubjectTest.java
```diff
@@ -114,5 +130,27 @@ public void testSubjectUserIdCanBeUpdated() {
...
     }
 
+    @Test
+    public void testAnonymisesUserIdentifiers() {
+        Subject subject = getSubject();
+        subject.setUserId("uid-test");
+        subject.setDomainUserId("duid-test");
+        subject.setNetworkUserId("nuid-test");
+        subject.setIpAddress("127.0.0.1");
+        assertNull(subject.getSubject(true).get("uid"));
```
mscwilson, line 140: Question: why did you call `subject.getSubject()` for each assertion, instead of saving the `subject.getSubject()` map as a variable and then using that? Is it to avoid making a variable?

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/SubjectTest.java
```diff
@@ -114,5 +130,27 @@ public void testSubjectUserIdCanBeUpdated() {
...
+        assertNull(subject.getSubject(true).get("duid"));
+        assertNull(subject.getSubject(true).get("tnuid"));
+        assertNull(subject.getSubject(true).get("ip"));
+    }
+
+    // Helper Methods
+
+    private Subject getSubject() {
```
mscwilson, line 148: Nit: This could be a good opportunity to change the name of this helper so it's not the same as Subject `getSubject()`, I was briefly confused

mscwilson (approved):
LGMT.
Good idea to anonymise the Subject identifiers to match the JS tracker

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/SubjectTest.java
matus-tomlein: Good point, fixed!

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/SubjectTest.java
matus-tomlein: It's just my personal preference for writing tests – I try to avoid introducing variables in tests because it makes them slightly more difficult to read as one has to look for how the variables are intialized instead of reading the test from top to bottom. It doesn't make much difference in this case though, here it's mostly out of habit.

AlexBenny (approved):
LGTM!!
Maybe we can make explicit in the doc that the trackers resets the session when the user is anonymised?

PR #556, opened 2022-11-04
Expose public constructor for ConfigurationBundle (close #550)

### LINKED ISSUE #550: Expose public constructor for ConfigurationBundle
`ConfigurationBundle` is used when creating default configurations in a remote configuration setup. However, currently the only public constructor takes JSON objects with the configurations which is not very user friendly.

We should expose other constructors that take `NetworkConfiguration`, `TrackerConfiguraiton`, and other objects.

matus-tomlein:
Issue #550 

Expose a public constructor for ConfigurationBundle that accepts a namespace and network configuration. Other configurations can be assigned to the public properties of the bundle object, so I didn't think it was necessary to add them to the constructor.

This is still not ideal, because the `ConfigurationBundle` class is in the `internal` package of the tracker. However, moving it out of there would be a breaking change so we need to do that in a major version.


snowplowcla:

PR #567, opened 2022-12-01
Hardcode the anonymised session user ID (close #566)

### LINKED ISSUE #566: Hardcode the anonymised session user ID
**Describe the bug**
On some Android devices (mostly, if not all, Motorola) `new UUID(0,0)` can return invalid UUIDs

Is it possible to check the generated UUID here in the Android code, to see if it's valid: 

https://github.com/snowplow/snowplow-android-tracker/blob/e8465d7f81dc3d688286c2b3dd6892714e801746/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/session/Session.java#L206

And if it's not valid, they could return some hardcoded string value instead, that is at least some valid UUID.

Originally reported via a zendesk ticket

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**
Here are some examples of other developers having similar issues with Motorola devices.
https://forums.lenovo.com/t5/MOTOROLA-Android-Developer-Community/Possible-bug-in-Moto-G-JVM/m-p/5077939
https://github.com/google/gson/issues/2103

mscwilson:
For issue #566.

When userAnonymisation is on, one of the identifiers that is anonymised is the session user ID (`userId` in `client_session`). The existing userId is replaced with a null UUID (`00000000-0000-0000-0000-000000000000`), previously generated like this: `new UUID(0, 0).toString()`.

A bug in some Motorola phones caused a truncation of the full null UUID `00000000-0000-0000-0000-000000000000` to `0000-0000` when `toString()` is called on it. In our tracker, this meant that some events were going to the Bad events rows, as the session user ID was no longer a valid UUID.

This change hardcodes the null UUID as a string.

### DIFF
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/session/Session.java
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/internal/session/Session.java
@@ -203,7 +203,7 @@ public class Session {
         Map<String, Object> sessionCopy = new HashMap<>(sessionValues);
         sessionCopy.put(Parameters.SESSION_EVENT_INDEX, eventIndex);
         if (userAnonymisation) {
-            sessionCopy.put(Parameters.SESSION_USER_ID, new UUID(0, 0).toString());
+            sessionCopy.put(Parameters.SESSION_USER_ID, "00000000-0000-0000-0000-000000000000");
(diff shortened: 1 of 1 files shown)


matus-tomlein (approved):
LGTM!

Btw, don't forget to do the same in the Kotlin branch.

PR #570, opened 2022-12-13
Migrate internal classes to Kotlin (close #564)

### LINKED ISSUE #564: Migrate internal classes to Kotlin
Following on from issue #561 where the published classes were translated.

mscwilson:
All of the files have now been migrated to Kotlin, heading towards the version 5 release


snowplow-tracker/src/main/java/com/snowplowanalytics/core/utils/Util.kt
```diff
@@ -0,0 +1,362 @@
...
+        return Base64.encodeToString(string.toByteArray(), Base64.NO_WRAP)
+    }
+
+    /**
+     * Generates a random UUID for each event.
+     *
+     * @return a UUID string
+     */
+    @JvmStatic
+    val uUIDString: String
```
matus-tomlein, line 71: Nit: For things that generate a different value on each access, I prefer to use functions instead of properties. With properties, there is an expectation that the value is being held and remains consistent.

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/StateManager.kt
```diff
@@ -0,0 +1,187 @@
...
+import com.snowplowanalytics.snowplow.event.Event
+import com.snowplowanalytics.snowplow.payload.SelfDescribingJson
+import com.snowplowanalytics.snowplow.tracker.InspectableEvent
+import java.util.*
+
+class StateManager {
+    
+    private val identifierToStateMachine = HashMap<String, StateMachineInterface>()
+    private val stateMachineToIdentifier = HashMap<StateMachineInterface, String>()
+    private val eventSchemaToStateMachine = HashMap<String?, MutableList<StateMachineInterface>>()
```
matus-tomlein, line 13: Just adding a note as I came across this file – there is no need to have the schema key an nullable string, I think we can remove the question mark here as well in other places in the file that deal with schemas and the schema registry. Generally, the less nullables we have, the better (so we don't have to handle them and to make the code easier to reason about).

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/Tracker.kt
```diff
@@ -0,0 +1,828 @@
...
+     * by adding in extra information to the payload:
+     * - The event contexts (limited to identify the device and app)
+     * - The Tracker Subject
+     * - The Tracker parameters
+     *
+     * @param payload Payload the raw event payload to be
+     * decorated.
+     * @param contexts The raw context list
+     */
+    private fun addServiceEventPayload(
```
matus-tomlein, line 678: I think we can remove this function, it is not used from anywhere and it is a private member so pretty safe to drop it. Probably just some relic of the past.

.github/workflows/build.yml
```diff
@@ -10,9 +10,9 @@ jobs:
       fail-fast: false
       matrix:
         api-level:
-          - 21 
           - 24
           - 30
+#          - 32
```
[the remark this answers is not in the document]
mscwilson: I tried to, but the Run Tests part of the build action failed with "Warning: Failed to find package 'system-images;android-32;google_apis;x86'". Perhaps the reactivecircus/android-emulator-runner action doesn't support API 32? I will try bumping its version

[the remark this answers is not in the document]
matus-tomlein: I just checked on the master branch and we don't test in CI using the API version 32 there, so I guess this commented out line was introduced in this PR? In that case, we don't have to address it here (we can look into the API 32 in some other issue), I just thought we are removing the version from testing.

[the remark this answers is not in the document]
mscwilson: I introduced it when I dropped API 21 from the test matrix, so there would still be 3 versions under test

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/ServiceProvider.kt
```diff
@@ -0,0 +1,342 @@
...
+
+    override fun orMakeEmitter(): Emitter {
+        return emitter ?: makeEmitter().also { emitter = it }
+    }
+    
+    override fun orMakeTracker(): Tracker {
+        return tracker ?: makeTracker().also { tracker = it }
+    }
+
+    override fun orMakeTrackerController(): TrackerControllerImpl {
```
matus-tomlein, line 184: Could we rename these `orMake...()` functions to `getOrMake...()` as they were before? I guess the `get` part was removed in the translation but the function name doesn't make sense without it.

.github/workflows/build.yml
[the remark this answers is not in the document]
matus-tomlein: Oh, I see, then how about trying version 31? Or just remove altogether, nobody said there need to be 3 :D

snowplow-tracker/build.gradle
```diff
@@ -16,15 +16,14 @@ project.ext {
...
     buildToolsVersion "30.0.2"
 
     useLibrary 'android.test.base'
 
     defaultConfig {
-        // With API 24: Remove `androidx.appcompat:appcompat` as it's needed for Consumer<T>.
-        minSdkVersion 21
-        targetSdkVersion 31
+        minSdkVersion 24
+        targetSdkVersion 32
```
matus-tomlein, line 26: I am a bit worried about updating this because the last time we did it, some customers weren't able to make use of the tracker because their apps targeted earlier SDK versions. The difference between compile and target SDKs is quite confusing to me but I think we should be more conservative about updating both of them, I would keep both at 31 unless there is a reason we need 32.

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/event/ApplicationInstallTest.kt
```diff
@@ -0,0 +1,75 @@
+package com.snowplowanalytics.snowplow.event
```
matus-tomlein, line 1: Btw, let's make a task for adding missing copyright headers in the tracker files – they are missing quite randomly. Also we should update the year in copyrights.

matus-tomlein (approved):
LGTM! Huuuuuge piece of work, well done!

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/event/ApplicationInstallTest.kt
mscwilson: I've made an issue for it

PR #572, opened 2022-12-15
Add configuration to send requests with user ID to a Focal Meter endpoint (close #571)

### LINKED ISSUE #571: Add configuration to send requests with user ID to a Focal Meter endpoint
This will integrate the tracker with the FocalMeter system by Kantar that measures audience of content through a router meter.

The tracker should make a request with the user ID from the session context whenever that changes. The URL endpoint to send the requests to should be configurable.

matus-tomlein:
Issue #571 

Implements integration with the FocalMeter system by Kantar. It is configured using the `FocalMeterConfiguration` class which has a single property – the Kantar endpoint to send the requests to.

The implementation uses a state machine (`FocalMeterStateMachine`) that checks the user ID in the client session context entity in events in the `entities(event, state)` function. Whenever that changes from the last seen value, it makes a request to the Kantar endpoint. It doesn't persist the information whether a request was already made, so after restarting the app (or creating a new tracker), it will make the request again.

Using the `entities(event, state)` function in state machine is not the ideal place for this check. It would be better if we had a dedicated API in the state machines to make these kinds of checks on tracked events. We plan to implement this in the v5 of the trackers, so we can refactor the implementation at that point.

## Documentation: Sending session user identifier to Kantar FocalMeter

The tracker has the ability to send the user identifier (`userId` present in the session context) to a [Kantar FocalMeter](https://www.virtualmeter.co.uk/focalmeter) endpoint. This integration enables measuring the audience of content through the FocalMeter router meter.

To enable this feature, you can pass the `FocalMeterConfiguration` configuration with the URL of the Kantar endpoint. For example:

```java
FocalMeterConfiguration focalMeterConfiguration = new FocalMeterConfiguration("https://thekantarendpoint.com");
Snowplow.createTracker(getApplicationContext(),
        namespace,
        networkConfiguration,
        focalMeterConfiguration
```


AlexBenny (approved):
👍 LGTM!

matus-tomlein:
Closing this PR since we need a different implementation (based on plugins) for the v5 tracker.

PR #576, opened 2023-01-31
Refactor event interface and rename contexts to entities (close #573)

### LINKED ISSUE #573: Refactor event interface and rename contexts to entities
Refactor the `InspectableEvent` interface to remove the `state` property and `addPayloadValues` which were used only internally in state machines. Having the `state` property in the published event interface required us to also publish various other state classes that were only internal.

Additionally, there is inconsistent naming of context entities in the tracker – they are called `contexts` in events and `entities` in state machines. Since entities better reflects our more recent thinking about how to name them, we should adopt that. We can keep the `contexts` properties in events but deprecate them in favor of `entities`.

[Here is](https://github.com/snowplow/snowplow-objc-tracker/issues/757) the same issue on the iOS tracker.

matus-tomlein:
Issue #573

This PR refactors the `InspectableEvent` interface to remove the `state` property and `addPayloadValues` which were used only internally in state machines. Having the `state` property in the published event interface required us to also publish various other state classes that were only internal.

I extracted the `state` and `addPayloadValues` to another interface – `StateMachineEvent` which is only internal within the tracker, not published. The `TrackerEvent` class which implemented `InspectableEvent` now also implements the `StateMachineEvent`.

Additionally, there is inconsistent naming of context entities in the tracker – they are called `contexts` in events and `entities` in state machines. Since entities better reflects our more recent thinking about how to name them, I adopted that. I kept the `contexts` properties in events but deprecated them in favor of `entities`.

### DIFF
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/core/statemachine/DeepLinkState.kt
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/core/statemachine/DeepLinkState.kt
@@ -10,7 +10,7 @@
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
  */
-package com.snowplowanalytics.core.tracker
+package com.snowplowanalytics.core.statemachine
 
 class DeepLinkState(val url: String, val referrer: String?) : State {
     var readyForOutput = false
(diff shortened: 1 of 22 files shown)


snowplow-tracker/src/main/java/com/snowplowanalytics/core/statemachine/DeepLinkState.kt
```diff
@@ -10,7 +10,7 @@
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
  */
-package com.snowplowanalytics.core.tracker
+package com.snowplowanalytics.core.statemachine
```
mscwilson, line 13: Having the state stuff within its own folder is much neater!

mscwilson (dismissed):
LGTM!

PR #577, opened 2023-01-31
Add ability to provide custom tracker plugins to inspect and enrich tracked events (close #574)

### LINKED ISSUE #574: Add ability to provide custom tracker plugins to inspect and enrich tracked events
Provide a way for users to implement custom plugins that can serve two goals (we may add more later):

1. Enrich tracked events with additional context entities.
2. Get notified about events after they are tracked and be able to inspect them including their context entities.

We could make use of state machines as the underlying mechanism to implement the functionality. The API should be based on configuration as controllers similar to other tracker APIs.

[Here is](https://github.com/snowplow/snowplow-objc-tracker/issues/750) the same issue on the iOS tracker.

matus-tomlein:
Issue #574 

Implements the ability to provide custom plugins that can intercept tracked events. Plugins provide callbacks that can enrich events with additional entities and inspect tracked events.

**Code snippet**

The API is explained in this snippet:

```kt
// identifier needs uniquely identify the plugin
val plugin = PluginConfiguration("myPlugin")

// entities closure can return context entities to enrich events
// the list of schemas to call the closure for is optional (it will be called for all events if null)
plugin.entities(listOf("iglu:...")) {
    return [
        SelfDescribingJson(schema: "iglu:xx", andData: ["yy": true])
    ]
}

// after track callback called on a background thread to inspect final tracked events
// one can also supply a list of schemas to call the closure only for specific events
plugin.afterTrack {
    print("Tracked event with $event.entities.size entities")
}

// the plugin is supplied to the tracker as a configuration
val tracker = Snowplow.createTracker(
    context = context,
    namespace = "ns",
    network = networkConfig,
    plugin
)

// one can inspect the enabled plugins and get the list of their identifiers
val pluginIdentifiers = tracker.plugins.identifiers

// add additional plugins (using PluginConfiguration as above)
tracker.plugins.addPlugin(otherPlugin)

// remove registered plugins by their identifiers
tracker.plugins.removePlugin("myPlugin")
```

**Implemented using state machines**

On the background, plugins are implemented using `PluginStateMachine` state machine. Each plugin has a separate state machine with the same identifier. The state machine calls the plugin closures in the `entities()` and `afterTrack()` functions.

**After track callback**

In order to implement the `afterTrack()` callback (which will be useful in plugins such as the one for FocalMeter, see other PR), I had to add this API to the state machines. It is called on a background thread to enable the plugins to make API requests without blocking the tracking code.

**Global context**

I changed the internal implementation of global contexts so that it is now built using plugins. This simplifies the code since we don't have deal with global contexts as a special case in `Tracker`, it is just another plugin/state machine.

The implementation was straightforward, since global context just adds entities based on some conditions which can be done in the `entities` closure in plugins.

There were no public API changes required to global context (from outside, it looks the same).

**Other refactoring in `Tracker`**

Finally, I tried to tidy up the code in the `Tracker.transformEvent()` and `Tracker.payload()` functions to make it more clear what is the sequence of the processing steps when events are tracked.

### DIFF
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/core/statemachine/PluginStateMachine.kt
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/core/statemachine/PluginStateMachine.kt
@@ -47,7 +47,7 @@ class PluginStateMachine(
         return null
     }
 
-    override fun entities(event: InspectableEvent, state: State?): List<SelfDescribingJson>? {
+    override fun entities(event: InspectableEvent, state: State?): List<SelfDescribingJson> {
         return entitiesConfiguration?.closure?.apply(event) ?: emptyList()
     }
 
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/core/statemachine/PluginStateMachine.kt
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/core/statemachine/PluginStateMachine.kt
@@ -1,0 +1,61 @@
+/*
+ * Copyright (c) 2015-2022 Snowplow Analytics Ltd. All rights reserved.
+ *
+ * This program is licensed to you under the Apache License Version 2.0,
+ * and you may not use this file except in compliance with the Apache License Version 2.0.
+ * You may obtain a copy of the Apache License Version 2.0 at http://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the Apache License Version 2.0 is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
+ */
+package com.snowplowanalytics.core.statemachine
+
+import com.snowplowanalytics.snowplow.configuration.PluginAfterTrackConfiguration
+import com.snowplowanalytics.snowplow.configuration.PluginEntitiesConfiguration
+import com.snowplowanalytics.snowplow.event.Event
+import com.snowplowanalytics.snowplow.payload.SelfDescribingJson
+import com.snowplowanalytics.snowplow.tracker.InspectableEvent
+import java.util.*
+
+class PluginStateMachine(
+    override val identifier: String,
+    val entitiesConfiguration: PluginEntitiesConfiguration?,
+    val afterTrackConfiguration: PluginAfterTrackConfiguration?
+) : StateMachineInterface {
+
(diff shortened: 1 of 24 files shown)


snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/Tracker.kt
```diff
@@ -481,74 +476,69 @@ class Tracker(emitter: Emitter, val namespace: String, var appId: String, contex
...
+        }
+    }
+
+    private fun addStateMachinePayloadValues(event: TrackerEvent) {
+        // Payload can be optionally updated with values based on internal state
+        stateManager.addPayloadValuesToEvent(event)
+    }
+
     private fun addBasicPropertiesToPayload(payload: Payload, event: TrackerEvent) {
+        // Event ID
```
mscwilson, line 524: good idea to break this method up with some explanations

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/TrackerEvent.kt
```diff
@@ -69,4 +72,48 @@ class TrackerEvent @JvmOverloads constructor(event: Event, state: TrackerStateSn
         }
         return result
     }
+
+    fun addContextEntity(entity: SelfDescribingJson) {
+        entities.add(entity)
+    }
+
+    fun wrapContextsToPayload(payload: Payload, base64Encoded: Boolean) {
```
mscwilson, line 80: nit: this could be `wrapEntitiesToPayload()`

mscwilson (approved):
LGTM! 💥

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/TrackerEvent.kt
matus-tomlein: Good point, fixed!

snowplow-tracker/src/main/java/com/snowplowanalytics/core/statemachine/PluginStateMachine.kt
```diff
@@ -0,0 +1,61 @@
...
+        get() {
+            val config = afterTrackConfiguration ?: return emptyList()
+            return config.schemas ?: Collections.singletonList("*")
+        }
+
+    override fun transition(event: Event, state: State?): State? {
+        return null
+    }
+
+    override fun entities(event: InspectableEvent, state: State?): List<SelfDescribingJson>? {
```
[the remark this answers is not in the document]
matus-tomlein: Fixed!

PR #579, opened 2023-02-02
Update copyrights to 2023

mscwilson:
Issue #578 

Adds copyright notices to files missing them, and updates all to 2023.

PR #580, opened 2023-02-02
Release/5.0.0 beta.1

mscwilson:
This is the second pre-release of the upcoming version 5 of the Android tracker written in Kotlin. This beta release brings new features as well as refactoring the tracker internals. All of the code is now in Kotlin.

The headline feature is the ability to provide custom tracker plugins to intercept tracked events. Plugins provide callbacks that can enrich events with additional entities and inspect tracked events. See this snippet for a glimpse of a tracker plugin:

```kt
// identifier needs to uniquely identify the plugin
val plugin = PluginConfiguration("myPlugin")

// entities closure can return context entities to enrich events
// the list of schemas to call the closure for is optional (it will be called for all events if null)
plugin.entities(listOf("iglu:...")) {
    return [
        SelfDescribingJson(schema: "iglu:xx", andData: ["yy": true])
    ]
}

// after track callback called on a background thread to inspect final tracked events
// one can also supply a list of schemas to call the closure only for specific events
plugin.afterTrack {
    print("Tracked event with $event.entities.size entities")
}

// the plugin is supplied to the tracker as a configuration
val tracker = Snowplow.createTracker(
    context = context,
    namespace = "ns",
    network = networkConfig,
    plugin
)

// one can inspect the enabled plugins and get the list of their identifiers
val pluginIdentifiers = tracker.plugins.identifiers

// add additional plugins (using PluginConfiguration as above)
tracker.plugins.addPlugin(otherPlugin)

// remove registered plugins by their identifiers
tracker.plugins.removePlugin("myPlugin")
```

**Enhancements**

* Add ability to provide custom tracker plugins to inspect and enrich tracked events (#574)

**Under the hood**

* Update copyrights to 2023 (#579)
* Refactor event interface and rename contexts to entities (#573)
* Migrate internal classes to Kotlin (#564)


CHANGELOG
```diff
@@ -1,6 +1,9 @@
-Version 5.0.0-alpha.1 (2022-11-30)
+Version 5.0.0-beta.1 (2022-02-02)
```
matus-tomlein, line 1: On the Swift side, I kept all the pre-releases as separate items in the changelog, but I think this approach sounds good too, we can overwrite them and replace with the stable release when we get to it. Just adding a note here to sync that this is the way we will go about it, don't have any problem with it.

mscwilson: Actually the V5.0.0-alpha.1 section had got in twice by accident, during the v4.1.1 release

PR #582, opened 2023-02-10
Fix crash with null pointer when accessing session (close #581)

### LINKED ISSUE #581: Fix crash with null pointer when accessing session
I recently upgraded to 4.1.1 from version 3.0.0 and there seems to be a crash reported from the sdk:

```
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'java.util.Map ee0.e.d()' on a null object reference
       at com.snowplowanalytics.snowplow.internal.session.Session.getSessionContext(Session.java:202)
       at com.snowplowanalytics.snowplow.internal.tracker.Tracker.workaroundForIncoherentSessionContext(Tracker.java:757)
       at com.snowplowanalytics.snowplow.internal.tracker.Tracker.track(Tracker.java:629)
       at com.snowplowanalytics.snowplow.internal.tracker.TrackerControllerImpl.track(TrackerControllerImpl.java:94)
```

i set the up the tracker as under:

```
Snowplow.setup(
            application,
            remoteConfig,
            listOf(configBundle)
        ) { configurationPair ->
            Snowplow.getDefaultTracker()?.attachGlobalContextToTracker() 
            -- manually attach global config since this is not possible at the moment with remoteConfig.
        }
```

and send the event as below:

```
Snowplow.getDefaultTracker()?.track(selfDescribingEvent)
```

Is there something i might be missing? Or may i please know why this crash is happening? I am confident this never happened with version 3.0.0

matus-tomlein:
Issue #581 

## Null pointer in `Session.getSessionContext`

I was able to replicate the null pointer exception reported in the issue in case session tracking is paused and no events have been tracked yet. In that case, the `state` internal variable is not initialised before being accessed. This can occur in remote configuration when the tracker re-created after retrieving a new configuration.

I created a test case that tested this scenario and fixed it by not adding the session context entity if session tracking is paused and state not initialised.

## Null pointer `SessionController.getSessionId`

This can happen when accessing the session ID before the session state is created. We should return a nullable string from the function, but that would be a breaking change here (we do that in v4). So instead I decided to just return an empty string in case the state is null.

### DIFF
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/internal/tracker/ConfigurationTest.java
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/internal/tracker/ConfigurationTest.java
@@ -113,6 +113,15 @@ public class ConfigurationTest {
     }
 
     @Test
+    public void sessionIdEmptyBeforeSessionInitialized() {
+        Context context = InstrumentationRegistry.getInstrumentation().getTargetContext();
+        NetworkConfiguration networkConfig = new NetworkConfiguration("fake-url", HttpMethod.POST);
+        TrackerController tracker = Snowplow.createTracker(context, "emptySessionIdTest", networkConfig);
+        String sessionId = tracker.getSession().getSessionId();
+        assertEquals("", sessionId);
+    }
+
+    @Test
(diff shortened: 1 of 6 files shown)


mscwilson (approved):
LGTM!

PR #583, opened 2023-02-15
Release/4.1.2

matus-tomlein:
This patch release fixes a bug that caused null pointer exceptions when accessing session information before it was initialised or when tracking events when session was not yet initialised.

**Bug fixes**

* Fix crash with null pointer when accessing session (#581)


mscwilson (approved):
LGTM 🚢🚀

PR #586, opened 2023-03-17
Add Jetpack Compose demo app

mscwilson:
Demonstrating how to instrument the tracker within a Compose app.


snowplow-demo-new/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,24 @@
...
+        android:icon="@mipmap/ic_launcher"
+        android:label="@string/app_name"
+        android:roundIcon="@mipmap/ic_launcher_round"
+        android:supportsRtl="true"
+        android:theme="@style/Theme.Snowplowandroidtracker">
+        <activity
+            android:name=".MainActivity"
+            android:exported="true"
+            android:label="@string/app_name"
+            android:theme="@style/Theme.Snowplowandroidtracker">
```
matus-tomlein, line 20: I needed to set this activity as the default activity to be launched, otherwise Android Studio complained that there is no launch activity. I copied this from the Kotlin demo:

```suggestion
            android:theme="@style/Theme.Snowplowandroidtracker">
            <intent-filter>
                <action android:name="android.intent.action.MAIN" />
                <category android:name="android.intent.category.LAUNCHER" />
            </intent-filter>
```

build.gradle
```diff
@@ -8,12 +11,13 @@ buildscript {
         }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.2.2'
+        
+        classpath 'com.android.tools.build:gradle:7.2.2'
```
matus-tomlein, line 15: Did you have to upgrade the Gradle version in order to make the demo work? I was planning to do this anyway for v5, but we should probably do it in a separate PR.

snowplow-demo-new/src/main/java/com/snowplowanalytics/snowplow_demo_new/ui/SchemaDetailScreen.kt
```diff
@@ -0,0 +1,146 @@
...
+    ) { contentPadding ->
+        Box(modifier = Modifier.padding(contentPadding)) {
+            Column(modifier = Modifier.padding(all = 8.dp)) {
+                Spacer(modifier = Modifier.width(40.dp))
+
+                Column(modifier = Modifier.verticalScroll(rememberScrollState())) {
+                    Text(
+                        text = "URL",
+                        modifier = Modifier.padding(4.dp),
+                        style = TextStyle(fontSize = 12.sp)
```
matus-tomlein, line 70: Nit: we could make use of the [styles in Material design](https://developer.android.com/jetpack/compose/designsystems/material#typography) instead of defining everything ourselves.

```suggestion
                        style = MaterialTheme.typography.overline
```

snowplow-demo-new/src/main/java/com/snowplowanalytics/snowplow_demo_new/ui/SchemaDetailScreen.kt
```diff
@@ -0,0 +1,146 @@
...
+    
+    val entity = SelfDescribingJson(
+        "iglu:com.snowplowanalytics.iglu/anything-a/jsonschema/1-0-0", 
+        hashMapOf("name" to schemaParts.name, "vendor" to schemaParts.vendor)
+    )
+    val event = ScreenView("detail", UUID.randomUUID()).entities(listOf(entity))
+
+    // Tracks a ScreenView with attached context entity.
+    // This entity records information about the specific schema being viewed
+    Tracking.tracker()?.track(event)
```
matus-tomlein, line 44: Could we use a `DisposableEffect` as [suggested here](https://stackoverflow.com/a/69464673) to track the screen view? Instead, I think this will track the event on every re-render of the component which might be more often than every screen.

snowplow-demo-compose/src/main/java/com/snowplowanalytics/snowplowdemocompose/data/Tracking.kt
```diff
@@ -0,0 +1,45 @@
...
+            emitterConfig
+        )
+    }
+
+    @Composable
+    fun TrackScreenView(screenName: String, 
+                        screenId: UUID? = UUID.randomUUID(), 
+                        entities: List<SelfDescribingJson>? = null,
+    ) {
+        LaunchedEffect(Unit, block = {
```
matus-tomlein, line 40: I think `DisposableEffect` is the preferred callback here. [See this official docs](https://developer.android.com/jetpack/compose/side-effects#disposableeffect) that also talk about the use case of tracking analytics data. Might be better to listen for the `ON_START` lifecycle event as they suggest.

Issue #585, opened 2023-03-15
Keep minSdkVersion lower for 5.x


Anigif:
**Is your feature request related to a problem? Please describe.**
We can see that for version 5.x, `minSdkVersion` is bumped to 24. This means (as far as we can see) that we won't be able to update, as the `minSdkVersion` of our app is currently 23. This means we either have to bump `minSdkVersion` of our app to 24 or not be able to update to 5.x (and possibly miss bugfixes and new features).

There might be reasons to why this bump was needed, maybe you'll be able to enlighten us in the reasoning? I tried to search a bit in the issues and commits, but couldn't see anything specific.

**Describe the solution you'd like**
The best solution for us at the moment would be to lower `minSdkVersion` to at least 23 for version 5.x, as we would also like to start using that.

**Describe alternatives you've considered**
I'm not sure what support plans there are for 4.x, but if `minSdkVersion` cannot be lowered any further,

matus-tomlein:
Thanks for the feedback @Anigif! The reason for bumping the min SDK version was that we adopted a few Java 8 features (in particular lambda functions) that seemed not to be supported in the API 23.

We will investigate again if there is a way to make this work even with API 23. If you have some suggestion how to do that, please shout!

hakonschia:
Hi @matus-tomlein, I'm a colleague of @Anigif.

I'm not sure if this could solve the issue, but have you considered using [desugaring](https://developer.android.com/studio/write/java8-support-table)? As far as I understand, desugaring allows you to use certain Java 8 features without restrictions on API level, so it might be something to look into.

Although I seem to be able to build the `next` branch when I set `minSdkVersion` back to 21 (as it was previously) without adding desugaring. Not sure if there's something I'm missing here, but it doesn't seem like a necessary bump? I changed it in all three modules and can build and run `snowplow-demo-java` and `snowplow-demo-kotlin`, and all the tests in `snowplow-android-tracker` are passing

matus-tomlein:
Thanks @hakonschia for the tip, that's very useful! I have managed to get it to work on SDK version 23 when using desugaring.

The `next` branch builds even without desugaring but the problems for me arised when I tried to run the `snowplow-demo-java` and track some events. The demo is creating a `GlobalContextsConfiguration` which on the background now uses the Java 8 lambda functions.

We'll make a PR to roll back bumping the `minSdkVersion`.

hakonschia:
Great to hear! Thanks for the quick replies

[...]

PR #588, opened 2023-03-23
Upgrade Gradle to 7.6 and lower minSdkVersion requirement to 21 (close #587 and #585)

### LINKED ISSUE #587: Upgrade Gradle to 7.6 and handle breaking changes
Upgrade Gradle from version 4 to version 7. Along with the change, upgrade all dependencies in the tracker such as the Android Gradle Plugin.

Upgrading Gradle brings several breaking changes especially in publishing. We should no longer use the `android-maven-gradle-plugin` and `gradle-maven-publish-plugin` which are no longer maintained. Instead, we should make use of the `maven-publsih` that is provided along with the Android Gradle Plugin.

matus-tomlein:
* Downgrade minSdkVersion to 21 and compileSdkVersion to 33 (#585)
   * Add desugaring for Java 8 features on older SDKs
* Upgrade Gradle to 7.6.1 (#587)
   * Update buildToolsVersion to 30.0.3
   * Update Kotlin version to 1.8
* Update publishing
   * Remove `com.github.dcendents:android-maven-gradle-plugin` and `com.vanniktech:gradle-maven-publish-plugin`
   * Make use of maven-publish provided with AGP
* Rename namespace of Java demo app to distinguish from the Kotlin demo
* Upgrade tracker, demo app and test dependencies


mscwilson (approved):
LGTM
Great to replace that deprecated mavenPublish task finally

PR #590, opened 2023-03-28
Reintroduce customContexts as a Java field and return a mutable list from entities in events

matus-tomlein:
The `customContexts` property in the `AbstractEvent` class was previously available as a public Java field. In the translation to Kotlin, it has changed into a property which changes how it can be used in Java. This PR changes it to a `@JvmField` and deprecates it in favour of the `entities` property.

The PR also changes the `entities` property to return a mutable list so that users can append to it. I also added a test to check that all the various ways to add entities work and added tracking an entity in the demo apps.


mscwilson (approved):
LGTM
Not sure what's happening with the commits but the Reintroduce customContexts etc one looks good

PR #591, opened 2023-03-29
Use Dokka for API docs

mscwilson:
We previously used Doxygen for API docs. Doxygen doesn't support Kotlin, so we need to move to Dokka, which builds off KDoc and/or Javadoc comments.

This PR updates the existing published API docs for KDoc, as well as adding overall polish. The core packages (non-published) have not been touched.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -2,7 +2,7 @@ name: Documentation
 
 on:
   push:
-#    branches: [master]
+    branches: [master]
 
 jobs:
   docs:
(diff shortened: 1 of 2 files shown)


matus-tomlein (commented):
Looks great!

Could we remove the internal core packages from the docs? We don't want users to use them and it's just unnecessary noise. I found there is a `suppress` option in the Dokka config ([see here](https://kotlin.github.io/dokka/1.4.10.2/user_guide/gradle/usage/#configuration-options)):

```
        perPackageOption {
            prefix = "com.snowplowanalytics.core" // will all sub-packages
            suppress = true
        }
```

matus-tomlein (approved):
LGTM!

PR #594, opened 2023-04-04
Release/5.0.0

mscwilson:


CHANGELOG
```diff
@@ -1,7 +1,30 @@
+Version 5.0.0 (2023-04-04)
```
matus-tomlein, line 1: Just for the sake of keeping full history, could we add version 4.1.2 changes here?

matus-tomlein (commented):
There is a problem with merging the release into master, because in the meantime we released a few patch v4 releases. I think what we'll need to do is to put the current commits on master to a `v4` branch. Then we can rollback the master commits back to the point where the work on the `release/5.0.0` started and then we should be able to merge this branch in. There might be a different way to achieve this, but basically we should back up the current master into another branch and then replace the master branch with this one.

matus-tomlein (approved):
LGTM! 🚀 🚢

Issue #598, opened 2023-05-06
Track new properties in platform context version 1-0-3 and make it configurable which properties to track


matus-tomlein:
Add the new properties in platform context entity (`mobile_context` schema) version 1-0-3:

Property | Description
--|--
isPortrait | A Boolean indicating whether the device orientation is portrait (either upright or upside down)
resolution | Screen resolution in pixels. Arrives in the form of WIDTHxHEIGHT (e.g., 1200x900). Doesn't change when device orientation changes
scale | Scale factor used to convert logical coordinates to device coordinates of the screen (uses UIScreen.scale on iOS and DisplayMetrics.density on Android)
language | System language currently used on the device (ISO 639)
appSetId | Android vendor ID scoped to the set of apps published under the same Google Play developer account (see https://developer.android.com/training/articles/app-set-id)
appSetIdScope | Scope of the `appSetId`. Can be scoped to the app or to a developer account on an app store (all apps from the same developer on the same device will have the same ID)

Also make it configurable which properties should be tracked in the entity in case the user does not want to track all of them.

**Related issues**

* #569
* #565
* #450

PR #600, opened 2023-05-11
Release/5.1.0

matus-tomlein:
This minor release extends the platform context entity tracked in events with new information:

* `isPortrait` – whether the device orientation is portrait (either upright or upside down).
* `resolution` – Screen resolution in pixels.
* `scale` – Scale factor used to convert logical coordinates to device coordinates of the screen.
* `language` – System language currently used on the device (ISO 639)
* `appSetId` – Android vendor ID scoped to the set of apps published under the same Google Play developer account
* `appSetIdScope` – Scope of the `appSetId`

It also makes it configurable which properties should be tracked in the platform context entity using the `TrackerConfiguration.platformContextEntities` configuration.

**Enhancements**

* Track new properties in platform context version 1-0-3 and make it configurable which properties to track (#598)


mscwilson (approved):
LGTM!

PR #602, opened 2023-05-19
Track install referrer details entity along with the application install event if available (close #249)

### LINKED ISSUE #249: Add tracking of Play Store referer

matus-tomlein:
Issue #249 

1. Refactors how the application install event is tracked – a deprecated `AsyncTask` was used before so I replaced with a background task using the `Executor`.
2. Adds the `referrer_details` entity to the install event in case the required `com.android.installreferrer:installreferrer` dependency is added in the project. The entity contains information about where the app was installed from and the install timestamps.


mscwilson (approved):
LGTM

PR #604, opened 2023-05-23
Add version to default remote configuration and don't update unless remote configuration is newer (close #603)

### LINKED ISSUE #603: Add version to default remote configuration and don't update unless remote configuration is newer
When the user configures the tracker using remote configuration and sets a default configuration, the default configuration will always be replaced by the fetched configuration regardless of whether it is different or not. This causes a new session to start (since a new tracker is created).

To prevent replacing the tracker in case the remote configuration is the same as the default configuration, we should add a configuration version to the default configuration and only replace it if the fetched configuration has a higher version.

[The same issue on the Android tracker.](https://github.com/snowplow/snowplow-ios-tracker/issues/779)

matus-tomlein:
Issue #603

Adds an option to specify the configuration version for default configuration in remote config. This version is then used to compare the default to the fetched configuration and only update it in case the fetched one has a newer version.

The benefit is that the default tracker is not recreated from the fetched configuration if it is the same. This also means that the session is not reset which used to happen when replacing the default configuration previously.


mscwilson (approved):
LGTM! - but wrong base branch

PR #609, opened 2023-05-29
Add a filter API to plugins to decide whether to track an event or not (close #608)

### LINKED ISSUE #608: Add a filter API to plugins to decide whether to track an event or not
Add an option to provide a `filter` callback in plugins that would return true or false deciding whether the event should be tracked or not. In case any plugin returns false, the event will be discarded and not tracked.

This will, for instance, enable filtering automatically tracked events such as from screen view autotracking and enable users to decide which ones to track.

matus-tomlein:
Issue #608 

Adds a new `filter` callback to plugins which enables deciding whether to track an event or not. This can be useful for filtering auto-tracked events by the tracker.

For instance, this snippet filters screen view events to keep only ones with a specific name property:

```kotlin
plugin.filter(
    schemas = listOf(
        "iglu:com.snowplowanalytics.snowplow/screen_view/jsonschema/1-0-0", // screen view events
    )
) { event ->
    event.payload["name"] == "Home Screen"
}
```

## Documentation

[Here is a commit adding the documentation for the filter function](https://github.com/snowplow/documentation/commit/79410621b205fdbbcc2ef3dd20264121e15a8238).


mscwilson (approved):
LGTM

PR #611, opened 2023-06-06
Add Snowplow ecommerce events and entities

mscwilson:
Implements the same ecommerce tracking APIs as in the JS/browser tracker. There is a matching PR for the iOS tracker [here](https://github.com/snowplow/snowplow-ios-tracker/pull/800).

There is one API change - `setPageType` to `setScreenType`. It makes more sense as a method, but could be confusing since it's still the Page entity. Might change it back to PageType...
Also, one API addition - the ability to remove set PageType and User entities.

There is a PR for the updated Snowplow docs [here](https://github.com/snowplow/documentation/pull/477).


snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/ecommerce/entities/Product.kt
```diff
@@ -0,0 +1,84 @@
...
+data class Product(
+    /**
+     * The SKU or product ID
+     */
+    val id: String,
+
+    /**
+     * The name or title of the product
+     */
+    val name: String? = null,
```
matus-tomlein, line 27: Maybe we should keep the properties mutable (`var`) so that users can initialize them also later instead of just in the constructor if they prefer to?

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/ServiceProvider.kt
```diff
@@ -48,6 +49,7 @@ class ServiceProvider(
     private var subjectController: SubjectControllerImpl? = null
     private var sessionController: SessionControllerImpl? = null
     private var gdprController: GdprControllerImpl? = null
+    private var ecommerceController: EcommerceControllerImpl? = null
```
matus-tomlein, line 52: This is a bit of a weird Java style dance that we do here with the `getOrMakeXY` and `makeXY` methods in addition to the `XY` property. I think we can achieve the same thing by making the property with lazy initialization – see the `pluginsController` below.

snowplow-tracker/src/main/java/com/snowplowanalytics/core/ecommerce/EcommerceEvent.kt
```diff
@@ -0,0 +1,80 @@
...
+import com.snowplowanalytics.core.constants.TrackerConstants
+import com.snowplowanalytics.snowplow.ecommerce.entities.Product
+import com.snowplowanalytics.snowplow.ecommerce.entities.Promotion
+import com.snowplowanalytics.snowplow.payload.SelfDescribingJson
+
+/**
+ * Interface for creating Ecommerce entities.
+ */
+interface EcommerceEvent {
+    fun productToSdj(product: Product) : SelfDescribingJson {
```
matus-tomlein, line 25: Suggestion: maybe the serialization could be implemented in the Product class? I think that would  better encapsulate it. As an example, [here is how I did similar serialization in the media APIs](https://github.com/snowplow/snowplow-android-tracker/blob/issue/606-media/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/media/entity/MediaPlayerEntity.kt#L64-L84).

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/event/AbstractEvent.kt
```diff
@@ -58,12 +61,24 @@ abstract class AbstractEvent : Event {
      * @return the event custom context entities
      */
     override val entities: MutableList<SelfDescribingJson>
-        get() = customContexts
+        get() {
+            if (isProcessing) {
+                if (entitiesForProcessing != null) {
```
matus-tomlein, line 66: Nit: just realized now that since we are calling `entitiesForProcessing` here twice (this line and the next), it results in the entities list being generated twice. Maybe we could avoid that (e.g., store in a local variable or use `?.let`)?

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/ecommerce/events/AddToCart.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * List of product(s) that were added to the cart.
+     */
+    var products: List<Product>,
+
+    /**
+     * The total value of the cart after this interaction.
+     */
+    var totalValue: Number,
```
matus-tomlein, line 40: Nit: I think it would be nicer to have a `Cart` class that encapsulates the `totalValue`, `currency` and `cartId` properties similar like the `Product` class. At least to me it would make it more explicit what the properties refer to in this event and also in the `RemoveFromCart` one.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/ecommerce/events/CheckoutStep.kt
```diff
@@ -0,0 +1,107 @@
...
+ * @param billingFullAddress Full billing address.
+ * @param deliveryProvider Can be used to discern delivery providers e.g. DHL, PostNL etc.
+ * @param deliveryMethod Store pickup, standard delivery, express delivery, international, etc.
+ * @param couponCode Coupon applied at checkout.
+ * @param accountType Type of account used on checkout, e.g. existing user, guest.
+ * @param paymentMethod Any kind of payment method the user selected to proceed. Card, PayPal, Alipay etc.
+ * @param proofOfPayment E.g. invoice or receipt.
+ * @param marketingOptIn If opted in to marketing campaigns to the email address.
+ */
+class CheckoutStep @JvmOverloads constructor(
```
matus-tomlein, line 37: Nit: The name `CheckoutStep` does not immediately make it clear whether this is an event or an entity. In the mobile media APIs, I decided to add the `Event` suffix to all event classes and `Entity` to the end of all entity classes. So my suggestion would be to name this `CheckoutStepEvent` to make it clear that it can be tracked. We could also rename `ProductEntity` but up to your consideration (I used the entity suffix in the media APIs mostly to avoid clashes with some Android API names such as `MediaPlayer`).

snowplow-tracker/src/main/java/com/snowplowanalytics/core/ecommerce/EcommerceControllerImpl.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.snowplowanalytics.snowplow.ecommerce.EcommerceController
+import com.snowplowanalytics.snowplow.ecommerce.entities.Product
+import com.snowplowanalytics.snowplow.ecommerce.entities.Promotion
+import com.snowplowanalytics.snowplow.payload.SelfDescribingJson
+
+@RestrictTo(RestrictTo.Scope.LIBRARY)
+class EcommerceControllerImpl(val serviceProvider: ServiceProviderInterface) : EcommerceController {
+
+    override fun setEcommerceScreen(type: String, language: String?, locale: String?) {
+        val plugin = PluginConfiguration("ecommercePageTypePluginInternal")
```
matus-tomlein, line 29: Should this entity be added to all events or just the `snowplow_ecommerce_action`? You could use the filter parameter in the `PluginConfiguration` to provide a list of schemas to which the entity should be added.

(the same comment for the `setEcommerceUser`).

matus-tomlein (approved):
LGTM, very nice work!

snowplow-tracker/src/main/java/com/snowplowanalytics/core/ecommerce/EcommerceControllerImpl.kt
mscwilson: I agree that it is surprising that it's added to all the events, not just the ecommerce ones. However, this is the behaviour in the JS tracker so I think it's best to keep it the same.

matus-tomlein (approved):

PR #614, opened 2023-06-13
Use default configuration for properties that are not configured using remote configuration (close #613)

### LINKED ISSUE #613: Use default configuration for properties that are not configured using remote configuration
The default configuration values should be maintained in case they are not overriden by the properties from a fetched remote configuration.

For example, if one configures a default tracker configuration with the following properties: `appId`, `platformContextProperties`, and `userAnonymisation`. And the remote configuration only changes the `appId`, the `platformContextProperties` and `userAnonymisation` properties should be kept from the default configuration.

iOS tracker issue: https://github.com/snowplow/snowplow-ios-tracker/issues/785

matus-tomlein:
Issue #613 

The goal of this PR is to make the fetched or cached remote configuration use properties from the default configuration in case they are not overriden. For example, if the default config sets the `appId` in tracker configuration, but the fetched remote configuration doesn't change it, we should still use the app ID from the default configuraiton.

To achieve this, I decided to remove the configuration update classes. Before, each configuration had a corresponding update class (e.g., `TrackerConfigurationUpdate`) which purpose was to keep track of changes made to the configuration after the tracker was initialized. It remembered which properties were changed and made sure that changes are kept even if the configuration is replaced from the remote endpoint.

Instead of the update classes, I moved the logic to check for the changes directly into the configuration classes. So the `TrackerConfiguration` class now has the ability to have a `sourceConfig` tracker configuration and when a property is requested, first it is checked if it is available in the current tracker configuration and, if not, it is retrieved from the `sourceConfig` (this is exactly what the update class was doing before).

Having the logic directly in the configuration classes, makes it possible to have multiple levels of the configurations. In particular, we now have the following levels:

1. The `ServiceProvider` has the top-most configuration. Properties of this are prioritized, but if they are not set, we go to the second level.
2. On this level we may have the fetched or cached remote configuration. If a property is missing here, we go to the third level.
3. This level is the default configuration. This is the final place where a property is looked for.

All these changes are internal and should not change the public API.


mscwilson (approved):
LGTM!

PR #617, opened 2023-06-14
Truncate URL scheme for page_url and page_refr properties (close #616)

### LINKED ISSUE #616: Truncate URL scheme for page_url and page_refr properties
The issue was originally reported in the React Native tracker: https://github.com/snowplow/snowplow-react-native-tracker/issues/190
Copying the issue description here:

**Describe the bug**
screen_view and deep_link_received events fail validation in the enricher on page_urlscheme (atomic schema) if the url scheme of a deep link is larger than [16 characters](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/atomic/jsonschema/1-0-0#L136). Deep links can have a url scheme that is longer than 16 characters (e.g. `someappwithaverylongscheme://...`).

I can work around it by truncating the url scheme before logging, but ideally I don't lose this data.

**To Reproduce**
Track an event via a deep link (e.g. a `trackDeepLinkReceivedEvent()`) with a url that has a urlscheme longer than 16 characters. 

**Expected behavior**
These events should pass validation.

**Smartphone (please complete the following information):**
 - Device: iPhone/Android devices
 - OS: iOS/Android
 - Version: snowplow-react-native-tracker 1.4.0

**Additional context**
I believe this workaround ([iOS](https://github.com/snowplow/snowplow-objc-tracker/blob/ec912a566073615ebcc9dc3a4c6b96a1cc6ed7e7/Snowplow/Internal/Tracker/SPTracker.m#L637  
)/[Android](https://github.com/snowplow/snowplow-objc-tracker/blob/ec912a566073615ebcc9dc3a4c6b96a1cc6ed7e7/Snowplow/Internal/Tracker/SPTracker.m#L637  
)) causes it.

The same issue on the iOS tracker: https://github.com/snowplow/snowplow-ios-tracker/issues/793

matus-tomlein:
Issue #616

This PR addresses the issue that the URL scheme may be quite long for deep links but the atomic events schema limits the max length of the scheme to 16 characters for page_url and page_refr properties, [see here](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/atomic/jsonschema/1-0-0#L136). This can lead to validation errors when deep links with longer URL schemes are tracked.

To solve the issue, the tracker now truncates the URL scheme in page_url and page_refr properties to 16 chars. This only affects the values in the atomic properties, the URLs in the deep link event and context entity are not truncated.

### DIFF
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/event/DeepLinkReceivedTest.kt
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/event/DeepLinkReceivedTest.kt
@@ -54,8 +54,8 @@ class DeepLinkReceivedTest {
         val context = InstrumentationRegistry.getInstrumentation().targetContext
 
         // Prepare DeepLinkReceived event
-        val event = DeepLinkReceived("url")
-            .referrer("referrer")
+        val event = DeepLinkReceived("someappwithaverylongscheme://url")
+            .referrer("someappwithaverylongscheme://referrer")
 
         // Setup tracker
         val trackerConfiguration = TrackerConfiguration("appId")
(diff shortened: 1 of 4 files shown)


mscwilson (approved):
LGTM!

PR #622, opened 2023-06-29
Truncate language in platform context entity to max 8 characters (close #621)

### LINKED ISSUE #621: Truncate language in platform context entity to max 8 characters
The schema for the `mobile_context` entity has validation for the `language` property for max 8 characters. This is causing validation errors for some languages such as `zh-Hans-CN`.

We should truncate the value to 8 characters on the tracker to avoid bad events.

Same issue on the iOS tracker: https://github.com/snowplow/snowplow-ios-tracker/issues/795

matus-tomlein:
Issue #621

This PR handles situations where the language property in the `mobile_context` entity was longer than 8 characters, which caused validation errors as the property has a `maxLength` setting of 8 chars.


mscwilson (approved):
LGTM!

Issue #629, opened 2023-07-24
Correctly add ecommerce entity in demo remote config


mscwilson:
The ecommerce entity addition isn't currently working, it needs to be part of the `setup` callback.

Issue #628, opened 2023-07-24
Prevent crash when Locale returns invalid language code


mscwilson:
**Describe the bug**
An Android device is returning an invalid, i.e. not part of ISO-639, language code from `Locale`. This is causing a crash during `DeviceInfoMonitor` initialisation when the platform context entity values are determined.

**To Reproduce**
Set the Locale to an unassigned code, e.g. "dk", then create a `PlatformContext`.

**Device information (please complete the following information):**
 - Device: Lenovo Tab M10 FHD Plus
 - OS: Android 10

PR #630, opened 2023-07-24
Release/5.4.1

mscwilson:
This is a patch release that fixes a crash resulting from Java's `Locale` util returning an inappropriate result, namely an invalid ISO-639 language code.

**Bug fixes**
Prevent crash when Locale returns invalid language code (#628)
Correctly add ecommerce entity in demo remote config (#629)


snowplow-demo-java/src/main/java/com/snowplowanalytics/snowplowtrackerdemojava/Demo.java
```diff
@@ -255,6 +255,21 @@ private boolean setupWithRemoteConfig(@NonNull Consumer<Boolean> callbackTracker
...
+
+            Objects.requireNonNull(Snowplow.getDefaultTracker()).getEcommerce().setEcommerceScreen(
+                    new EcommerceScreenEntity(
+                            "demo_app_screen"
+
+                    )
+            );
+
+            SelfDescribingJson exampleGlobalEntity = new SelfDescribingJson(
+                    "iglu:com.snowplowanalytics.iglu/anything-a/jsonschema/1-0-0",
```
mfolcini, line 267: can't this be added to a constant and reused everywhere?

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/internal/tracker/PlatformContextTest.kt
```diff
@@ -246,6 +246,25 @@ class PlatformContextTest {
         Assert.assertEquals("12345678", sdjData[Parameters.MOBILE_LANGUAGE])
     }
 
+    @Test
+    fun catchesInvalidLocaleLanguageNoMocking() {
```
mfolcini, line 250: I don't understand this test, if you are putting an invalid locale it´s dropping an exception, I thought the change was to return null if invalid? (catch (e: MissingResourceException) {
                null
            }) ?

mscwilson: I wrote the test to replicate/find a bug, and didn't update it once I'd added the fix

mfolcini (approved):

Issue #637, opened 2023-09-05
Fix invalid consent document context entity in case no name or description is given


matus-tomlein:
When the consent granted or withdrawn events are tracked with a consent document that doesn't have any name or description, the consent document entity is tracked with null name and null description. [According to the schema](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/consent_document/jsonschema/1-0-0), this is invalid because the properties can't be null. However, they are not required, so we can omit adding them to the entity.

PR #638, opened 2023-09-05
Release/5.4.4

matus-tomlein:
This patch release fixes invalid consent document context entity that was sent in case no name or description was given.

**Bug fixes**

* Fix invalid consent document context entity in case no name or description is given (#637)


mscwilson (approved):
LGTM

PR #640, opened 2023-09-08
Add API to decorate link with user/session info (close #639)

### LINKED ISSUE #639: Add API to decorate link with user/session info

greg-el:
This PR adds a method to `TrackerController` that decorates a URI, provided by the user, with session/user info for cross-device/platform tracking.

For example, calling decorateLink on `appSchema://path/to/page` will return:
`appSchema://path/to/page?_sp={domainUserId}.{timestamp}.{sessionId}.{subjectUserId}.{sourceId}.{platform}.{reason}`

### DIFF
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/LinkDecoratorTest.kt
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/LinkDecoratorTest.kt
@@ -1,0 +1,168 @@
+package com.snowplowanalytics.snowplow.tracker
+
+
+import android.net.Uri
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import com.snowplowanalytics.core.utils.Util.urlSafeBase64Encode
+import com.snowplowanalytics.snowplow.Snowplow
+import com.snowplowanalytics.snowplow.configuration.NetworkConfiguration
+import com.snowplowanalytics.snowplow.configuration.SessionConfiguration
+import com.snowplowanalytics.snowplow.configuration.SubjectConfiguration
+import com.snowplowanalytics.snowplow.configuration.TrackerConfiguration
+import com.snowplowanalytics.snowplow.controller.SessionController
+import com.snowplowanalytics.snowplow.controller.TrackerController
+import com.snowplowanalytics.snowplow.event.ScreenView
+import com.snowplowanalytics.snowplow.network.HttpMethod
+import com.snowplowanalytics.snowplow.util.TimeMeasure
+import org.junit.Assert
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.util.concurrent.TimeUnit
+
+
+@RunWith(AndroidJUnit4::class)
+class LinkDecoratorTest {
+    private lateinit var tracker: TrackerController
+    private lateinit var session: SessionController
+    private lateinit var userId: String
+    private lateinit var appId: String
+    private val subjectUserId = "subjectUserId"
+    private val subjectUserIdEncoded = urlSafeBase64Encode(subjectUserId)
+    private val testLink = Uri.parse("http://example.com")
+    private val epoch = "\\d{13}"
+
+    private fun matches(pattern: String, result: Uri) {
+        val regex = Regex("^${pattern.replace(".", "\\.").replace("?", "\\?")}$")
+        Assert.assertTrue(
+            "$result\ndoes not match expected:  $pattern", regex.matches(result.toString())
+        )
+    }
+
+
+    @Before
+    fun before() {
+        tracker = getTracker()
+        session = tracker.session!!
+        userId = session.userId
+        appId = urlSafeBase64Encode(tracker.appId)
+    }
+
+    @Test
+    fun testWithoutSession() {
+        val tracker = getTrackerNoSession()
+        val result = tracker.decorateLink(testLink)
+        Assert.assertEquals(null, result)
+    }
+
+    @Test
+    fun testDecorateUriWithExistingSpParam() {
+        tracker.track(ScreenView("test"))
+
+        val pattern = "http://example.com?_sp=$userId.$epoch.${session.sessionId}..$appId"
+        val result =
+            tracker.decorateLink(testLink.buildUpon().appendQueryParameter("_sp", "test").build())
+
+        matches(pattern, result!!)
+    }
+
+    @Test
+    fun testDecorateUriWithOtherParam() {
+        tracker.track(ScreenView("test"))
+
+        val pattern = "http://example.com?a=b&_sp=$userId.$epoch.${session.sessionId}..$appId$"
+        val result =
+            tracker.decorateLink(testLink.buildUpon().appendQueryParameter("a", "b").build())
+
+        matches(pattern, result!!)
+    }
+
+    @Test
+    fun testDecorateUriWithParameters() {
+        tracker.track(ScreenView("test"))
+
+        val sessionId = session.sessionId
+        val decorate = { c: CrossDeviceParameterConfiguration -> tracker.decorateLink(testLink, c)!! }
+
+        matches(
+            "http://example.com?_sp=$userId.$epoch.$sessionId",
+            decorate(CrossDeviceParameterConfiguration(sourceId = false))
+        )
+
+        matches(
+            "http://example.com?_sp=$userId.$epoch.$sessionId..$appId",
+            decorate(CrossDeviceParameterConfiguration())
+        )
+
+        matches(
+            "http://example.com?_sp=$userId.$epoch.$sessionId..$appId.mob",
+            decorate(CrossDeviceParameterConfiguration(sourcePlatform = true))
+        )
+
+        matches(
+            "http://example.com?_sp=$userId.$epoch.$sessionId.$subjectUserIdEncoded.$appId.mob",
+            decorate(CrossDeviceParameterConfiguration(sourcePlatform = true, subjectUserId = true))
+        )
+
+        matches(
+            "http://example.com?_sp=$userId.$epoch.$sessionId...mob",
+            decorate(CrossDeviceParameterConfiguration(sourceId = false, sourcePlatform = true))
+        )
+
+        matches(
+            "http://example.com?_sp=$userId.$epoch..$subjectUserIdEncoded.$appId",
+            decorate(CrossDeviceParameterConfiguration(sessionId = false, subjectUserId = true))
+        )
+
+        matches(
+            "http://example.com?_sp=$userId.$epoch..$subjectUserIdEncoded.$appId",
+            decorate(CrossDeviceParameterConfiguration(sessionId = false, subjectUserId = true))
+        )
+
+
+        matches(
+            "http://example.com?_sp=$userId.$epoch",
+            decorate(CrossDeviceParameterConfiguration(sourceId = false, sessionId = false))
+        )
+    }
+
+    private fun getTracker(): TrackerController {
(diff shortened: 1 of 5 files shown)


mscwilson (approved):
LGTM

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/controller/TrackerController.kt
```diff
@@ -116,4 +120,23 @@ interface TrackerController : TrackerConfigurationInterface {
...
+     * @param uri The URI to add the query string to
+     * @param parameters The parameters to include in the query string (defaults to all)
+     *
+     * @return Optional Uri:
+     * - null if [Session.userId] is null from `sessionContext(false)` being passed in [TrackerConfiguration]
+     * - otherwise, decorated Uri
+     */
+    fun decorateLink(
+        uri: Uri,
+        parameters: List<CrossDeviceParameter> = CrossDeviceParameter.values().toList()
```
matus-tomlein, line 140: In the JS tracker, @igneel64 called this the `extendedCrossDomainLinker` and set the defaults to `sessionId` and `sourceId`. Would be great to do something similar here, maybe we could call the property `extendedParameters` (to make it clear that it these are not all the parameters being set in the URI) and by default only set the session and source ID?

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/controller/TrackerController.kt
```diff
@@ -116,4 +120,23 @@ interface TrackerController : TrackerConfigurationInterface {
...
+
+    /**
+     * Adds user and session information to a URI.
+     *
+     * For example, calling decorateLink on `appSchema://path/to/page` will return:
+     *
+     *      `appSchema://path/to/page?_sp=userId.timestamp.sessionId.appId.platform.domainUserId`
+     *
+     * @param uri The URI to add the query string to
+     * @param parameters The parameters to include in the query string (defaults to all)
```
matus-tomlein, line 132: Just a note to also make it clear that this only relates to the optional (extended) parameters, userID and timestamp are set regardless.

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/CrossDeviceParameter.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * Value of [Tracker.platform]
+     */
+    SOURCE_PLATFORM,
+
+    /**
+     * Value of [Subject.userId]
+     */
+    USER_ID,
```
matus-tomlein, line 28: I think we are missing the `reason` parameter here, which is the link text. That is a bit awkward one because we can't automatically populate based on the link as in the JS tracker, but will probably need to accept another optional string argument in the `decorateLink` call.

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/LinkDecoratorTest.kt
```diff
@@ -0,0 +1,179 @@
...
+            val result = tracker.decorateLink(testLink, param)
+
+            matchesRegex(pattern, result!!)
+        }
+    }
+
+    fun getTracker(): TrackerController {
+        val context = InstrumentationRegistry.getInstrumentation().targetContext
+
+        val networkConfiguration = NetworkConfiguration("fake-url", HttpMethod.POST)
```
matus-tomlein, line 142: We prefer to use the mocked network connection in tests so that we don't make unnecessary requests, see here: https://github.com/snowplow/snowplow-android-tracker/blob/master/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/PluginsTest.kt#L219

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/TrackerControllerImpl.kt
```diff
@@ -70,6 +71,42 @@ class TrackerControllerImpl  // Constructors
         return tracker.track(event)
     }
 
+
+    override fun decorateLink(uri: Uri, parameters: List<CrossDeviceParameter>): Uri? {
+        // UserId is a required parameter of `_sp`
+        if (this.session?.userId == null) {
```
matus-tomlein, line 77: Should we do some logging here? Maybe a debug log that the userId is null and so the link couldn't be decorated? Something like we do here: https://github.com/snowplow/snowplow-android-tracker/blob/master/snowplow-tracker/src/main/java/com/snowplowanalytics/core/session/SessionControllerImpl.kt#L46C13-L46C16

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/CrossDeviceParameter.kt
```diff
@@ -0,0 +1,29 @@
...
+enum class CrossDeviceParameter {
+    /**
+     * Value of [SessionController.sessionId]
+     */
+    SESSION_ID,
+
+    /**
+     * Value of [Tracker.appId]
+     */
+    SOURCE_ID,
```
matus-tomlein, line 18: Based on the JS tracker implementation, the `userId` should be before the `sourceId`.

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/TrackerControllerImpl.kt
```diff
@@ -70,6 +71,42 @@ class TrackerControllerImpl  // Constructors
...
+            return null
+        }
+
+        val values = hashMapOf(
+            CrossDeviceParameter.SESSION_ID to (this.session?.sessionId ?: ""),
+            CrossDeviceParameter.SOURCE_ID to this.appId,
+            CrossDeviceParameter.SOURCE_PLATFORM to this.devicePlatform.value,
+            CrossDeviceParameter.USER_ID to (this.subject.userId ?: "")
```
matus-tomlein, line 85: We will need to do a base64 encoding for the `userId`, `sourceId` and `reason` to make sure that they don't contain any dots which break the parsing on enrich.

matus-tomlein (commented):
Very nice implementation!

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/CrossDeviceParameter.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * Value of [Tracker.platform]
+     */
+    SOURCE_PLATFORM,
+
+    /**
+     * Value of [Subject.userId]
```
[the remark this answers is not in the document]
greg-el: Yeah I agree, the naming conventions of this are pretty confusing. I've renamed `USER_ID` to `SESSION_USER_ID` and tried to impove the docstring to show the mapping between the tracker values and url params.

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/CrossDeviceParameter.kt
greg-el: Added a `reason` arg to `decorateLink`

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/TrackerControllerImpl.kt
greg-el: Added a log.

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/TrackerControllerImpl.kt
```diff
@@ -70,6 +72,50 @@ class TrackerControllerImpl  // Constructors
...
+        if (this.session?.userId == null) {
+            Logger.track(TAG, "$uri could not be decorated as session.userId is null ")
+            return null
+        }
+
+        val values = hashMapOf(
+            CrossDeviceParameter.SESSION_ID to (this.session?.sessionId ?: ""),
+            CrossDeviceParameter.SOURCE_ID to base64Encode(this.appId),
+            CrossDeviceParameter.SOURCE_PLATFORM to this.devicePlatform.value,
+            CrossDeviceParameter.SUBJECT_USER_ID to base64Encode(this.subject.userId ?: "")
```
matus-tomlein, line 91: I think this `Util.base64Encode` function does not produce the URL-safe variant of base 64 encoding. To get the URL safe version (as done in the JS tracker), I think you will need to pass these options to the [Android Base64](https://developer.android.com/reference/android/util/Base64) class:

```kt
Base64.encodeToString(string.toByteArray(), Base64.URL_SAFE | Base64.NO_PADDING | Base64.NO_WRAP)
```

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/controller/TrackerController.kt
```diff
@@ -116,4 +122,48 @@ interface TrackerController : TrackerConfigurationInterface {
...
+     * - null if [Session.userId] is null from `sessionContext(false)` being passed in [TrackerConfiguration]
+     * - otherwise, decorated Uri
+     */
+    fun decorateLink(
+        uri: Uri,
+        extendedParameters: List<CrossDeviceParameter> = listOf(
+            CrossDeviceParameter.SOURCE_ID,
+            CrossDeviceParameter.SESSION_ID
+        ),
+        reason: String = String()
```
matus-tomlein, line 167: Nit: making this an optional string would be nicer (at least to my taste)

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/LinkDecoratorTest.kt
```diff
@@ -0,0 +1,199 @@
...
+
+        matchesRegex(pattern, result!!)
+    }
+
+    @Test
+    fun testDecorateUriWithNoParams() {
+        tracker.track(ScreenView("test"))
+
+        val pattern =
+            Regex("""http://example\.com\?_sp=$userId\.\d{13}\.${session.sessionId}\.\.$appId""")
```
matus-tomlein, line 77:
```suggestion
            Regex("""^http://example\.com\?_sp=$userId\.\d{13}\.${session.sessionId}\.\.$appId$""")
```

Nit: I think these regular expressions are a bit tricky because they don't test what comes after the appId, so we don't test that the trailing dots are trimmed. Would be nicer to add the `$` at the end to test that as well.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/controller/TrackerController.kt
```diff
@@ -116,4 +122,48 @@ interface TrackerController : TrackerConfigurationInterface {
...
+     *
+     *      `appSchema://path/to/page?_sp=domainUserId.timestamp.sessionId.subjectUserId.sourceId.platform.reason`
+     *
+     * Required (auto-filled by this method):
+     * - `domainUserId`: Value of [SessionController.userId]
+     * - `timestamp`: ms precision epoch timestamp
+     *
+     * Optional:
+     * - `sessionId`: Value of [SessionController.sessionId]
+     * - `userId`: Value of [Subject.userId]
```
matus-tomlein, line 139:
```suggestion
     * - `subjectUserId`: Value of [Subject.userId]
```

matus-tomlein (commented):
Just one concern about the URL-safe base64 encoding, but otherwise LGTM!

matus-tomlein (approved):
LGTM! 👍

matus-tomlein (approved):
Nice to see this being in line with the iOS tracker! Just one comment about moving the configuration out of the core module.

Issue #641, opened 2023-09-20
Never retry any of the failed requests


ismaelgsi:
We've integrated the Snowplow sdk on a project where we don't want to retry any of the failed requests.

We've configured the tracker so it doesn't retry any of the failed requests with HTTP error status codes, using customRetryForStatusCodes on the EmitterConfiguration. But we're still seeing that all the requests that failed because of a timeout or any other connection error are being saved to the database and the tracker tries to send them again.

It would be great to have a way to define on that EmitterConfiguration that we don't want to retry any of the failed requests. Something like:

`EmitterConfiguration().retryFailedRequests(false)`

Changing the implementation of the shouldRetry function on RequestResult class returning this value directly when it's set.

We want to avoid having traffic peaks with a lot of events being sent at the same time, for example after fixing an issue on the snowplow server. And right now there is no way to configure the tracker to do that.

Does this change make sense to you?

Thanks!!

PR #642, opened 2023-10-02
Release/5.5.0

matus-tomlein:
This release adds an option to disable retrying all failed requests to the collector. The option is called `EmitterConfiguration.retryFailedRequests`. If configured, events that fail to be sent in the first request to the collector will be dropped. This may be useful in situations where it's necessary to prevent traffic spikes with many events being sent at the same time.

**Enhancements**
Add option to disable retrying any failed requests (#641)


snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/configuration/EmitterConfiguration.kt
mscwilson, line 1: nit: add `retryFailedRequests: true` to the default values for the API docs at the top of the class

mscwilson (approved):
LGTM!

Issue #649, opened 2023-11-30
Change default buffer option to single


matus-tomlein:
Even though we state the default buffer option in the tracker is 10, in practice the buffer option is 1 (due to issue #648). This means that events are sent right when they are tracked.

This is also the configuration on the JavaScript tracker (buffer size 1) and is a good default to use in client side apps to prevent some events hanging in the event store when the app is quit/uninstalled earlier than the event store is cleared.

After fixing issue #648, in order to keep this behaviour, we will need to change the default buffer size to 1.

Issue #648, opened 2023-11-30
Flush events only when the buffer is full


matus-tomlein:
As reported in [this issue on the iOS tracker](https://github.com/snowplow/snowplow-ios-tracker/issues/827), the tracker is ignoring the buffer size when new events arrive. Regardless of what is the buffer size, the emitter triggers an emit on new events.

For more information, refer to the iOS tracker issue.

Issue #646, opened 2023-11-29
Make network requests serially in network connection


matus-tomlein:
**Background**
There are 2 settings in the emitter – buffer option (default 10) and emit range (default 150). Buffer option is the number of events that have to pile up before the tracker makes a request. Emit range is the number of events that can be sent in one iteration of the emitter (basically at once).

If due to collector being down, 200 events pile up in the event queue on the tracker, the emitter will make 15 requests with 10 events and fire them at the same time to the collector (because emit range is 150 and buffer size is 10 so max 10 events fit in one request => 15 requests).

**Problem**
The goal of making parallel requests is to send the events to the collector faster. However, after doing some tests, this does not seem to be the case in practice – sending 500 events in batches of 25 per request takes around the same total time to finish when making requests in parallel as it does when making requests serially.

Moreover, making requests in parallel increases the likelihood of duplicate events ending up in the warehouse. If there is a problem with the network connection while the requests are made, more events will be resent. Also making multiple requests at once increases the chance of network failures on weaker connections.

**Suggestion**
The suggested approach is to:
1. Make requests serially. Adapt the `NetworkConnection` to receive one request at a time.
2. Change the semantics of the buffer option and emit range so that buffer option is the minimum number of events that have to pile up before making a network request, and emit range is the maximum number of events that can be sent in a single request to the collector (may be larger than buffer option in case of events piling up in the event queue due to previous failed requests).

[Same issue on the iOS tracker](https://github.com/snowplow/snowplow-ios-tracker/issues/846).

PR #650, opened 2023-12-01
Revise the emitter behaviour

matus-tomlein:
This PR contains three changes:

### Flush events only when the buffer is full (#648)

Previously the buffer size configuration didn't work as intended – requests to the collector were initiated right after the event was tracked regardless of the buffer size.

This change fixes that and only makes the request once the buffer size is reached.

### Change default buffer option to single (#649)

As mentioned before, previously the buffer size was effectively equal to 1 (even though we stated it was 10). Fixing the above bug, we would now change the buffer size to 10.

Buffer size 1 is probably a better default for client side trackers to avoid losing events that don't make it to a request before an app is quit/uninstalled. Also the same is used on the JavaScript tracker.

This change sets the default buffer size to 1 (from a user's point of view there won't be a difference compared to previous versions).

### Make network requests serially in network connection (#646)

This change tries to address a problem with duplicate events often showing up in the warehouse. It is possible that this is called by sending too many parallel requests at once which causes either the requests to fail or timeout and then be retried.

After doing some tests, I discovered that making parallel requests from the client has very little benefit. The total time it takes to send all the events is very close to what it would be if requests were made serially.

The new behaviour updates how the `emitRange` configuration is used – it now tells how many events should be added to one request. For this reason, I also lowered the default value to 25.

When merging this PR, I will rebase the commits so that they show up individually in the release branch.

### DIFF
--- a/snowplow-demo-java/src/main/java/com/snowplowanalytics/snowplowtrackerdemojava/Demo.java
+++ b/snowplow-demo-java/src/main/java/com/snowplowanalytics/snowplowtrackerdemojava/Demo.java
@@ -296,7 +296,6 @@ public class Demo extends Activity implements LoggerDelegate {
         EmitterConfiguration emitterConfiguration = new EmitterConfiguration()
                 .requestCallback(getRequestCallback())
                 .threadPoolSize(20)
-                .emitRange(500)
                 .byteLimitPost(52000);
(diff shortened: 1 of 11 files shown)


mscwilson (approved):
LGTM
Great to finally sort out the `bufferOption`/`sentLimit`/`emitRange` properties!

PR #655, opened 2023-12-08
Do not track the screen view again when app comes to foreground (close #653)

### LINKED ISSUE #653: Do not track the screen view again when app comes to foreground
Currently, when screen view autotracking is enabled and the app goes to background and comes back to foreground, another screen view event is tracked.

This is not the expected behaviour. Moving an app to background and back to foreground does not change the current screen view and shouldn't cause a new event to be tracked (except for background and foreground events).

We should update the screen view autotracking feature not to track new screen view events when the app is moved to foreground.

matus-tomlein:
Issue #653

Previously a screen view event was tracked again by the screen view autotracking feature when the app moved to foreground. This is not expected because the screen doesn't change when the app is in background and it is not consistent with how the screen view autotracking works on iOS.

This PR adds a check before autotracking screen views that checks that the new screen view is different from the last one.


mscwilson (approved):
LGTM

PR #661, opened 2024-01-19
Add configurable limit for the maximum age and number of events in the event store and remove old events before sending (close #660)

### LINKED ISSUE #660: Add configurable limit for the maximum age and number of events in the event store and remove old events before sending
The event store is currently unbounded meaning that if events fail to be sent to the collector, they are never removed from the event store. This can lead to it growing endlessly (if for instance an ad blocker blocks the collector domain) which is likely to have an impact on the app.

We should add configurable limits to the event store so that old events are removed. We could have two such limits:

* maximum event store size – in case the number of events surpasses this threshold, oldest events will be removed until we are under the threshold. Default could be 1000.
* maximum evnet age – in case there are events older than this threshold, we should remove them. Default could be 30 days.

matus-tomlein:
Issue #660 

The event store is currently unbounded meaning that if events fail to be sent to the collector, they are never removed from the event store. This can lead to it growing endlessly (if for instance an ad blocker blocks the collector domain) which is likely to have an impact on the app.

We should add configurable limits to the event store so that old events are removed. We could have two such limits:

* maximum event store size – in case the number of events surpasses this threshold, oldest events will be removed until we are under the threshold. Default could be 1000.
* maximum event age – in case there are events older than this threshold, we should remove them. Default could be 30 days.

The implementation in this PR adds these limits and enforces them before each emit attempt. They are enforced by running a single SQL statement which should be relatively efficient.

In addition, this PR removes some unnecessary optional types in the event store interface.

### DIFF
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/internal/emitter/storage/EventStoreTest.kt
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/internal/emitter/storage/EventStoreTest.kt
@@ -166,7 +166,7 @@ class EventStoreTest {
     @Throws(InterruptedException::class)
     fun testRemoveRangeOfEvents() {
         val eventStore = eventStore()
-        val idList: MutableList<Long?> = ArrayList()
+        val idList: MutableList<Long> = ArrayList()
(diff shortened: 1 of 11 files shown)


mscwilson (approved):
LGTM!

PR #663, opened 2024-01-24
Fix demo crash when using R8 full mode (close #652)

### LINKED ISSUE #652: Enabling R8 Full mode breaks the demo apps
**Describe the bug**
When R8 full mode is on, the demo code breaks. This is important cause from gradle version 8.0+ R8 full mode is on by default

**To Reproduce**
Ive created a demo [here](https://github.com/Andr3Carvalh0/snowplow-android-tracker). All you need to do is build the `release` variant of the snowplow-demo-compose

**Expected behavior**
Whether R8 full mode is on or off tracking should work

**Screenshots**
When I compile the demo with the `android.enableR8.fullMode` set to `true`, I get this error
Screenshot 2023-12-07 at 18 03 46

If I still minify the app, but disable R8 Full mode, the app works normally

**Device informatoin (please complete the following information):**
 - Device: Any
 - OS: Any
 - Version: Any
 
Only requirement is that you minify your code using R8 full mode

mscwilson:
For issue #652.

The crash in the Compose demo was caused by missing Retrofit classes when full R8 minification/shrinking is used. This PR adds ProGuard exemption rules for the relevant classes.

Also, dependencies for getting the appSetId and androidIdfa (AAID) platform context entity properties have been added to the Compose demo, along with ProGuard rules so that they still work with R8.

### DIFF
--- a/snowplow-demo-compose/proguard-rules.pro
+++ b/snowplow-demo-compose/proguard-rules.pro
@@ -19,3 +19,24 @@
 # If you keep the line number information, uncomment this to
 # hide the original source file name.
 #-renamesourcefileattribute SourceFile
+
+# Fixes R8 full mode bug
+# Also see https://github.com/square/retrofit/issues/3751
+-keep,allowobfuscation,allowshrinking interface retrofit2.Call
+-keep,allowobfuscation,allowshrinking class retrofit2.Response
+-keep,allowobfuscation,allowshrinking class kotlin.coroutines.Continuation
+
+## Please add these rules to your existing keep rules in order to suppress warnings.
+## This is generated automatically by the Android Gradle plugin.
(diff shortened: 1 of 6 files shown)


snowplow-demo-compose/proguard-rules.pro
```diff
@@ -18,4 +18,19 @@
 
 # If you keep the line number information, uncomment this to
 # hide the original source file name.
-#-renamesourcefileattribute SourceFile
\ No newline at end of file
+#-renamesourcefileattribute SourceFile
+
+# Fixes R8 full mode crash
```
matus-tomlein, line 23: We should also add these to the [setup section in the docs](https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/mobile-trackers/installation-and-set-up/android-tracker/).

matus-tomlein (approved):
LGTM!

PR #665, opened 2024-01-26
Undeprecate PageView event

mscwilson:
The `PageView` event class was accidentally marked as deprecated during the migration to Kotlin for v5. This change reverses that.

### DIFF
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/event/PageView.kt
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/event/PageView.kt
@@ -16,7 +16,7 @@ import com.snowplowanalytics.core.constants.Parameters
 import com.snowplowanalytics.core.constants.TrackerConstants
 
 /**
- * A pageView event.
+ * A PageView event. This event has been designed for web trackers, and is not suitable for mobile apps.
  * @param pageUrl The page URL.
  */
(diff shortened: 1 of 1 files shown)


snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/event/PageView.kt
```diff
@@ -18,9 +18,7 @@ import com.snowplowanalytics.core.constants.TrackerConstants
 /**
  * A pageView event.
  * @param pageUrl The page URL.
- * @Deprecated This event has been designed for web trackers, and is not suitable for mobile apps.
```
matus-tomlein, line 21: Maybe we can keep the comment? We do have a comment like that on the iOS tracker and it is true that page views are not suitable for mobile apps (disregarding web views), so might be a relevant info to guide users away from this in case they want standard mobile tracking? Android Studio doesn't seem to do anything about the comment.

matus-tomlein (approved):
LGTM!

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/event/PageView.kt
mscwilson: I'll move the comment to the main part of the docstring, instead of using the Deprecated annotation

PR #666, opened 2024-01-26
Tidy AbstractEvent internal properties

mscwilson:
Breaking changes in the base `AbstractEvent` class, to standardise the API with the iOS tracker.

### DIFF
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/event/AbstractEventTest.kt
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/event/AbstractEventTest.kt
@@ -24,14 +24,21 @@ class AbstractEventTest {
     @Test
     fun testAddsEntitiesUsingAllAPIs() {
         val event = ScreenView("screen")
-        val entity = SelfDescribingJson("schema3", "data3")
+        val entity1 = SelfDescribingJson("schema1", "data1")
+        val entity2 = SelfDescribingJson("schema2", "data2")
+        val entity3 = SelfDescribingJson("schema3", "data3")
 
-        event.entities.add(SelfDescribingJson("schema1", "data1"))
-        event.contexts(listOf(SelfDescribingJson("schema2", "data2")))
-        event.entities(listOf(entity))
-
+        event.entities.add(entity1)
         Assert.assertEquals(1, event.entities.count())
-        Assert.assertEquals(1, event.contexts.count())
-        Assert.assertTrue(event.entities.contains(entity))
+        
+        event.entities(listOf(entity2))
+        Assert.assertEquals(2, event.entities.count())
+        
+        event.contexts(listOf(entity3))
+        Assert.assertEquals(3, event.entities.count())
+        Assert.assertEquals(3, event.contexts.count())
+        Assert.assertTrue(event.entities.contains(entity1))
+        Assert.assertTrue(event.entities.contains(entity2))
+        Assert.assertTrue(event.entities.contains(entity3))
     }
 }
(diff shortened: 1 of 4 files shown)


snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/event/AbstractEventTest.kt
```diff
@@ -24,14 +24,14 @@ class AbstractEventTest {
     @Test
     fun testAddsEntitiesUsingAllAPIs() {
         val event = ScreenView("screen")
+        val entity = SelfDescribingJson("schema3", "data3")
 
         event.entities.add(SelfDescribingJson("schema1", "data1"))
-        event.customContexts.add(SelfDescribingJson("schema2", "data2"))
-        event.entities(listOf(SelfDescribingJson("schema3", "data3")))
-        event.contexts(listOf(SelfDescribingJson("schema4", "data4")))
+        event.contexts(listOf(SelfDescribingJson("schema2", "data2")))
```
matus-tomlein, line 30: We are not testing the outcome of these two calls – should we have asserts here? Better to swap their order so that we test that the `add()` adds to the existing entities.

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/event/AbstractEvent.kt
```diff
@@ -38,15 +55,15 @@ abstract class AbstractEvent : Event {
     
     // Builder methods
 
-    /** Adds a list of context entities to the existing ones. */
-    fun entities(entities: List<SelfDescribingJson>?): AbstractEvent {
-        entities?.let { customContexts.addAll(entities) }
+    /** Replace the context entities attached to the event with a new list of entities. */
+    fun entities(entities: List<SelfDescribingJson>): AbstractEvent {
+        this.entities = entities.toMutableList()
```
matus-tomlein, line 60: I am a bit of a coward and have a bad feeling about this 😅 On one hand, this looks like a builder function which usually replace the values, but going from a behaviour that is non-destructive (appending) to a destructive one (replacing) is a nasty breaking change. Nobody will notice this in the migration guide and it may lead to some very hard to find bugs with missing data.

I think I would prefer if we kept the Android behaviour and updated on iOS to append instead of replace. Other changes in this PR look good, I'd keep those. What do you think?

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/event/AbstractEvent.kt
```diff
@@ -22,10 +22,27 @@ import java.util.*
  * - "True" timestamp: user-defined custom event timestamp
  */
 abstract class AbstractEvent : Event {
-    /** List of custom contexts associated with the event.  */
-    @Deprecated("Old nomenclature.", ReplaceWith("entities"))
-    @JvmField
-    val customContexts: MutableList<SelfDescribingJson> = LinkedList()
+    private var _entities: MutableList<SelfDescribingJson> = LinkedList()
```
matus-tomlein, line 25: Nit: this looks slightly more Kotlin like:

```suggestion
    private var _entities = mutableListOf<SelfDescribingJson>()
```

snowplow-tracker/src/main/java/com/snowplowanalytics/snowplow/event/AbstractEvent.kt
mscwilson: That's fair! I do think the API makes more sense as full replacement. But being cautious is good.

matus-tomlein (approved):
LGTM@

PR #669, opened 2024-01-29
Update copyright notices

mscwilson:
Updates the copyright notices to `2015-present` so they'll no longer need updating annually.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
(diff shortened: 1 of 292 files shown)


LICENSE
```diff
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2023 Snowplow Analytics Ltd.
```
matus-tomlein, line 189: Same as on iOS tracker, could we make this 2015-present?

matus-tomlein (approved):

PR #670, opened 2024-01-31
Enable lifecycle autotracking by default (close #651)

### LINKED ISSUE #651: Enable lifecycle autotracking by default
Up to the v5 version of the tracker, lifecycle autotracking is not enabled by default. However, lifecycle events are crucial to understand the usage of the app and user engagement. It is not possible to calculate screen time without knowing whether the app is in foreground or background.

As we are introducing screen engagement tracking in the v6 version which will be enabled by default, we should also enable lifecycle tracking by default.

matus-tomlein:
Issue #651

Up to the v5 version of the tracker, lifecycle autotracking is not enabled by default. However, lifecycle events are crucial to understand the usage of the app and user engagement. It is not possible to calculate screen time without knowing whether the app is in foreground or background.

As we are introducing screen engagement tracking in the v6 version which will be enabled by default, we should also enable lifecycle tracking by default.

### DIFF
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/TrackerDefaults.kt
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/TrackerDefaults.kt
@@ -32,7 +32,7 @@ object TrackerDefaults {
     var applicationContext = true
     var exceptionAutotracking = true
     var diagnosticAutotracking = false
-    var lifecycleAutotracking = false
+    var lifecycleAutotracking = true
     var screenViewAutotracking = true
     var screenEngagementAutotracking = true
     var installAutotracking = true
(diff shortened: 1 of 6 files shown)


snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/TrackerDefaults.kt
```diff
@@ -32,7 +32,7 @@ object TrackerDefaults {
     var applicationContext = true
     var exceptionAutotracking = true
     var diagnosticAutotracking = false
-    var lifecycleAutotracking = false
+    var lifecycleAutotracking = true
```
mscwilson, line 35: while you're here, remove the unused `threadCount`?

mscwilson (approved):
LGTM!

PR #674, opened 2024-02-12
Fix wrong screen entity info assigned for screen end event and in case events are tracked right before screen view (close #673)

### LINKED ISSUE #673: Fix wrong screen entity info assigned for screen end event and in case events are tracked in parallel
Due to a bug in how the screen state is updated in the tracker, the screen end event is assigned the wrong information – instead of referring to the previous screen name and ID, it contains information of the subsequent screen.

This bug is not only limited to screen end events but may also appear in case an event is tracked right before a screen view event – the event will be assigned the next screen view instead of the one used when it was tracked.

matus-tomlein:
Issue #673 

Due to a bug in how the screen state is updated in the tracker, the screen end event is assigned the wrong information – instead of referring to the previous screen name and ID, it contains information of the subsequent screen.

This bug is not only limited to screen end events but may also appear in case an event is tracked right before a screen view event – the event will be assigned the next screen view instead of the one used when it was tracked.

### DIFF
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/screen/ScreenViewAutotrackingTest.kt
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/screen/ScreenViewAutotrackingTest.kt
@@ -61,10 +61,10 @@ class ScreenViewAutotrackingTest {
         NotificationCenter.postNotification("SnowplowScreenView", mapOf(
             "event" to ScreenView(name = "Screen2").activityClassName("Screen2")
         ))
-        Thread.sleep(2000)
+        Thread.sleep(500)
 
-        val numberOfScreenViews = eventSink.trackedEvents.filter { it.schema == TrackerConstants.SCHEMA_SCREEN_VIEW }
-        Assert.assertEquals(2, numberOfScreenViews.size)
+        val numberOfScreenViews = eventSink.trackedEvents.filter { it.schema == TrackerConstants.SCHEMA_SCREEN_VIEW }.size
+        Assert.assertEquals(2, numberOfScreenViews)
     }
 
     // --- PRIVATE
--- a/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/screen/ScreenViewAutotrackingTest.kt
+++ b/snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/screen/ScreenViewAutotrackingTest.kt
@@ -16,6 +16,7 @@ import android.content.Context
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.platform.app.InstrumentationRegistry
 import com.snowplowanalytics.core.constants.TrackerConstants
+import com.snowplowanalytics.core.emitter.Executor
 import com.snowplowanalytics.core.utils.NotificationCenter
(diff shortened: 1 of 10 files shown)


snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/screen/ScreenViewAutotrackingTest.kt
```diff
@@ -69,9 +61,10 @@ class ScreenViewAutotrackingTest {
         NotificationCenter.postNotification("SnowplowScreenView", mapOf(
             "event" to ScreenView(name = "Screen2").activityClassName("Screen2")
         ))
-        Thread.sleep(200)
+        Thread.sleep(2000)
 
-        Assert.assertEquals(2, numberOfScreenViews)
+        val numberOfScreenViews = eventSink.trackedEvents.filter { it.schema == TrackerConstants.SCHEMA_SCREEN_VIEW }
```
mscwilson, line 66: nit: strange name for a list, I was expecting an int

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/screenviews/ScreenSummaryStateMachineTest.kt
```diff
@@ -94,9 +94,14 @@ class ScreenSummaryStateMachineTest {
...
         Assert.assertEquals(3, events.size)
 
-        val screenSummary = getScreenSummary(events.find { it.schema == TrackerConstants.SCHEMA_SCREEN_END })
+        val screenEnd = events.find { it.schema == TrackerConstants.SCHEMA_SCREEN_END }
+        val screenSummary = getScreenSummary(screenEnd)
         Assert.assertEquals(10.0, screenSummary?.get("foreground_sec"))
         Assert.assertEquals(0.0, screenSummary?.get("background_sec"))
+
+        // should have the screen name of the first screen view
+        val screenEndScreen = screenEnd?.entities?.find { it.map["schema"] == TrackerConstants.SCHEMA_SCREEN }
```
mscwilson, line 103: I've read the word "screen" so much it doesn't look like a real word anymore

mscwilson (approved):
LGTM!

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/screen/ScreenViewAutotrackingTest.kt
matus-tomlein: Good catch, fixed now.

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/Tracker.kt
```diff
@@ -333,8 +328,9 @@ class Tracker(
         override fun apply(data: Map<String, Any>) {
             if (screenViewAutotracking) {
                 val event = data["event"] as? ScreenView?
-                event?.let { event ->
-                    getScreenState()?.let { state ->
+                if (event != null) {
```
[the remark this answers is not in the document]
matus-tomlein: It had a slightly different behaviour which I think was unintended – the `?: track(event)` part could be executed even in case the previous `let` block was evaluated but returned `null`. I think this if-else better reflects what should happen and is a bit more clear.

Issue #601, opened 2023-05-12
Using global context when creating a tracker causes crashes on API 23


frodstan:
**Describe the bug**
Initializing the sdk with a global context on Android 6 throws an exception. I have tried both with a static context and with context generator.

`FATAL EXCEPTION: main
                                                                                                    Process: com.example.snowplow_test, PID: 17105
                                                                                                    java.lang.NoClassDefFoundError: com.snowplowanalytics.core.globalcontexts.GlobalContextPluginConfiguration$$ExternalSyntheticLambda0
                                                                                                    	at com.snowplowanalytics.core.globalcontexts.GlobalContextPluginConfiguration.getEntitiesConfiguration(GlobalContextPluginConfiguration.kt:28)
                                                                                                    	at com.snowplowanalytics.snowplow.configuration.PluginConfigurationKt.toStateMachine(PluginConfiguration.kt:62)
                                                                                                    	at com.snowplowanalytics.core.tracker.ServiceProvider$makeTracker$builder$1.invoke(ServiceProvider.kt:293)
                                                                                                    	at com.snowplowanalytics.core.tracker.ServiceProvider$makeTracker$builder$1.invoke(ServiceProvider.kt:262)
                                                                                                    	at com.snowplowanalytics.core.tracker.Tracker.<init>(Tracker.kt:356)
                                                                                                    	at com.snowplowanalytics.core.tracker.ServiceProvider.makeTracker(ServiceProvider.kt:297)
                                                                                                    	at com.snowplowanalytics.core.tracker.ServiceProvider.getOrMakeTracker(ServiceProvider.kt:187)
                                                                                                    	at com.snowplowanalytics.core.tracker.ServiceProvider.<init>(ServiceProvider.kt:87)
                                                                                                    	at com.snowplowanalytics.snowplow.Snowplow.createTracker(Snowplow.kt:237)
                                                                                                    	at com.example.snowplow_test.MainActivity.initializeSnowplow(MainActivity.kt:68)
                                                                                                    	at com.example.snowplow_test.MainActivity.onCreate(MainActivity.kt:37)`

**To Reproduce**
To reproduce this issue, create a new Android app from template with minimum Android SDK 23.

Add snowplow library (version 5.0.0 - 5.1.0, 4.x.x has worked for us previously).

Initialize the sdk in MainActivity onCreate() with a global context sent to Snowplow.createTracker()
Pasting in from documentation here will cause the crash: 
[...]

PR #675, opened 2024-02-13
Proposal/Attempt to remove library desugaring

dpgmedia-gbero:
We observed crashes with devices running Android API 23. 
Initialization of FocalMeterConfiguration would crash.

A thread (https://github.com/snowplow/snowplow-android-tracker/issues/601) in the issue tracker suggested to use of coreLibraryDesugaring but we believe there might be a different option : Remove usage of java.util.function.* in order to get rid of library desugaring.

I'm not sure how to entirely test this but can this be looked at ?

Many thanks,

### DIFF
--- a/snowplow-demo-java/build.gradle
+++ b/snowplow-demo-java/build.gradle
@@ -34,7 +34,6 @@ android {
     }
 
     compileOptions {
-        coreLibraryDesugaringEnabled true
         sourceCompatibility = JavaVersion.VERSION_1_8
         targetCompatibility = JavaVersion.VERSION_1_8
     }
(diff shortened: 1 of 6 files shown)


snowplowcla:

dpgmedia-gbero:
I understand your concern. We decided to disable FocalMeterConfiguration on Android 6 and lower instead to circumvent this issue. You can close this or keep it for reference if you'd foresee to get rid of core library desugaring

plastiv:
Hi @matus-tomlein !

Could you consider to evaluate if `desugar_jdk_libs_minimal` can be used? Based on description it includes function package snowplow sdk has used. And if you agree to not extend the scope of java standard lib usage that may minimize the bundled dependency.

* https://github.com/google/desugar_jdk_libs/blob/master/CHANGELOG.md#version-200-2022-09-16

matus-tomlein:
Thank you for the suggestion @plastiv! That makes sense, I'll make sure we evaluate this.

Issue #662, opened 2024-01-24
Track percent progress information in the auto-tracked percent progress event in media tracking


matus-tomlein:
**Describe the bug**
Currently, the percent progress property of the event is missing.

**Expected behavior**
We should track the current percent progress that the event is fired for.

[Same issue on the iOS tracker](https://github.com/snowplow/snowplow-ios-tracker/issues/864)

PR #676, opened 2024-02-14
Release/6.0.1

mscwilson:
Two bug fixes. The media tracking PercentProgress event now includes the percent progress. Also, the screen state is now correctly updated, so that all events have the right screen details.

## Bug fixes
- Fix wrong screen entity info assigned for screen end event and in case events are tracked right before screen view (#673)
- Add percent progress to event (#662)


matus-tomlein (approved):
LGTM!

Issue #620, opened 2023-06-23
Snowplow.createTracker takes 10 seconds?


peterdk:
**Describe the bug**
When initiating Snowplow it takes 10 seconds to complete that call.....

**To Reproduce**
 val network = NetworkConfiguration(endpoint = "url", HttpMethod.POST)
 val config = TrackerConfiguration(context.packageName).screenViewAutotracking(true)
Snowplow.createTracker(
                context = context,
                namespace = "namespace",
                network = network,
                config,
            )

**Expected behavior**
200ms setup time or something reasonable.

**Device information (please complete the following information):**
 - Device: Pixel 6a
 - OS: Android 14
 - Version com.snowplowanalytics:snowplow-android-tracker:5.2.0

**Additional context**
Not useable this way

peterdk:
Also on Android 9, Samsung Note

matus-tomlein:
Hi @peterdk, that is very unexpected, could you please help me replicate this?

I have created a blank new Android app and pasted in the code that you provided above, [see here](https://github.com/matus-tomlein/snowplow-android-demo-test-lag/blob/main/app/src/main/java/com/example/testlagapp/MainActivity.kt#L42-L51).
However, when I try to test, the `createTracker` call is instant:

```
D/Snowplow: Fri Jun 23 12:58:45 GMT+02:00 2023 Snowplow tracker starting
I/TetheringManager: registerTetheringEventCallback:com.example.testlagapp
D/Snowplow: Fri Jun 23 12:58:45 GMT+02:00 2023 Snowplow tracker created
```

Could you please tell me what is different about your setup or help me extend the app to replicate the problem?

peterdk:
Yeah, I just moved the code to onCreate, like in the docs, and there it runs indeed very fast.
We use Koin with a lot of dependencies, so I'll look into this next week to see if I can find out what is causing this hang. Can it be that we call createTracker on a background thread and that gives issues?

matus-tomlein:
I don't think running on a background thread should be the problem - I have just tried it and the call was instant as well.

peterdk:
I debugged, and we indeed load our analytics module and thus also Snowplow, on a background coroutine dispatcher/thread.
I can reproduce the slow start if I make the main thread during that process busy:
```
CoroutineScope(Dispatchers.Default (  + SupervisorJob()).launch{ createTracker()}
Thread.sleep(5000)
```
[...]

PR #678, opened 2024-02-28
Release/6.0.2

matus-tomlein:
This release fixes a problem with fetching the Android Advertising ID (IDFA) that delayed the `createTracker` call to take 10 seconds in case it was run on a background thread. The tracker now doesn't retrieve any platform context properties during the `createTracker` call (they are fetched on background threads as events are processed), making it non-blocking to initialize the tracker.

**Bug fixes**

* Fix createTracker call hanging for 10 seconds if run on a background thread (#620)


mscwilson (approved):
LGTM!

PR #681, opened 2024-03-28
Catch more generic exceptions for InstallReferrer (close #647)

### LINKED ISSUE #647: Handle more generic exceptions when using InstallReferrer
**Describe the bug**
Currently The SnowPlow library is using this library as "compileOnly". 

compileOnly "com.android.installreferrer:installreferrer:2.2"

compileOnly means that it's mandatory for the library, but is not shipped with the artifact. This currently means that the SnowPlow library cannot be used in a project that doesn't have the installreferrer. And currently makes a crash:

`java.lang.NoSuchMethodError: No virtual method getGooglePlayInstantParam()Z in class Lcom/android/installreferrer/api/ReferrerDetails; or its super classes (declaration of 'com.android.installreferrer.api.ReferrerDetails' appears in /data/app/~~PHQ96RItx6M1UppnjKIAiQ==/com.inditex.zara-kj-XCASo7_iDxnet0B7pkw==/base.apk)
                                      	at pj1.f.onInstallReferrerSetupFinished(InstallReferrerDetails.kt:88)
                                      	at l8.a$a.onServiceConnected(InstallReferrerClientImpl.java:37)
                                      	at android.app.LoadedApk$ServiceDispatcher.doConnected(LoadedApk.java:2313)
                                      	at android.app.LoadedApk$ServiceDispatcher$RunConnection.run(LoadedApk.java:2350)
                                      	at android.os.Handler.handleCallback(Handler.java:958)
                                      	at android.os.Handler.dispatchMessage(Handler.java:99)
                                      	at android.os.Looper.loopOnce(Looper.java:230)
                                      	at android.os.Looper.loop(Looper.java:319)
                                      	at android.app.ActivityThread.main(ActivityThread.java:8893)
                                      	at java.lang.reflect.Method.invoke(Native Method)
                                      	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:608)
                                      	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1103)`

The crash happens in **InstallReferrerDetails.kt** line 72. There is a try/catch protecting it. But the catch is only handling RemoteException, while the system throws a NoSuchMethodError. NoSuchMethodError is not a subclass of RemoteException.

In order to allow using the library without installreferrer library, **I would capture all possible exceptions** (throwable: Throwable) in the try/catch.

**To Reproduce**
Use the last version of SnowPlow library in a project that doesn't have installreferrer library.
Then send some events.

**Expected behavior**
Should not crash when the app using the SnowPlow library doesn't have the installreferrer library

**Screenshots**
Not applicable

**Device informatoin (please complete the following information):**
 - Device: All devices [Tested in Samsung, Huawei]
 - OS: All android versions [Tested in 12, 13, 14]
 - Browser None
 - Version None

mscwilson:
For issue #647 

The `InstallReferrerDetails` class currently checks whether the InstallReferrer dependency is present or not, but if it is, then only catches `RemoteException`s. InstallReferrer below v1.1 doesn't have the `googlePlayInstantParam` property which causes a `NoSuchMethodError`.

Unfortunately the `googlePlayInstantParam` is the one required property for the `referrer_details` schema, so if InstallReferrer v1.0 is used it's not possible to add the entity at all.

This change catches exceptions generically, to include `NoSuchMethodError`.

### DIFF
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/InstallReferrerDetails.kt
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/InstallReferrerDetails.kt
@@ -14,7 +14,6 @@
 package com.snowplowanalytics.core.tracker
 
 import android.content.Context
-import android.os.RemoteException
 import com.android.installreferrer.api.InstallReferrerClient
 import com.android.installreferrer.api.InstallReferrerStateListener
 import com.android.installreferrer.api.InstallReferrerClient.InstallReferrerResponse
@@ -72,7 +71,7 @@ class InstallReferrerDetails(
                                     googlePlayInstantParam = response.googlePlayInstantParam
                                 )
                                 callback(referrer)
-                            } catch (_: RemoteException) {
+                            } catch (_: Throwable) {
                                 Logger.d(TAG, "Install referrer API remote exception.")
                                 callback(null)
                             }
@@ -99,7 +98,7 @@ class InstallReferrerDetails(
             })
         }
 
-        fun isInstallReferrerPackageAvailable(): Boolean {
+        private fun isInstallReferrerPackageAvailable(): Boolean {
             try {
                 Class.forName("com.android.installreferrer.api.InstallReferrerStateListener")
                 return true


snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/InstallReferrerDetails.kt
```diff
@@ -72,7 +71,7 @@ class InstallReferrerDetails(
                                     googlePlayInstantParam = response.googlePlayInstantParam
                                 )
                                 callback(referrer)
-                            } catch (_: RemoteException) {
+                            } catch (_: Throwable) {
```
matus-tomlein, line 74: This is fine, but I wonder if we should catch the `NoSuchMethodError` exception specifically to give a more concrete error message saying that the problem likely is that they need to update the InstallReferrer to v1.1?

VenomVendor:
As the code will always crash without the dependency, why was the decision to catch Exception instead of updating [transitive dependencies?](https://github.com/snowplow/snowplow-android-tracker/blob/c1be49c802bafb340b56a67b3ba599d9ea566197/snowplow-tracker/build.gradle#L93).

Can this PR also update about the [missing info in the doc](https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/mobile-trackers/installation-and-set-up/android-tracker/#installing)

mscwilson:
Hi @VenomVendor good shout about the docs!

I don't understand the first part of your message sorry. The code already caught when the dependency is completely absent, now it will also catch versions without the required functionality.

VenomVendor:
I meant, why to catch the exception if we can update transitive dependency.
[Here, we're using `compileOnly`](https://github.com/snowplow/snowplow-android-tracker/blob/c1be49c802bafb340b56a67b3ba599d9ea566197/snowplow-tracker/build.gradle#L93), instead if we could update to `implementation`. This exception handling can be removed, as the library with specific version will be bundled.

I was in an assumption that the dependency install-referrer is mandatory. Found this check `isInstallReferrerPackageAvailable()`.

Issue #657, opened 2023-12-18
Can't create a tracker with an Application context


eygraber:
**Describe the bug**
When calling `Snowplow.createTracker` with a `Context` retrieved from `Application.applicationContext`, sending an event logs an error, and no events are sent:

```
Tried to access visual service WindowManager from a non-visual Context:com.myapp.MyApp@50529e1 WindowManager should be accessed from Activity or other visual Context. Use an Activity or a Context created with Context#createWindowContext(int, Bundle), which are adjusted to the configuration and visual bounds of an area on screen.
                 java.lang.IllegalAccessException: Tried to access visual service WindowManager from a non-visual Context:com.myapp.MyApp@50529e1
                 	at android.app.ContextImpl.getSystemService(ContextImpl.java:2207)
                 	at android.content.ContextWrapper.getSystemService(ContextWrapper.java:934)
                 	at com.snowplowanalytics.core.tracker.Subject.setDefaultScreenResolution(Subject.kt:228)
                 	at com.snowplowanalytics.core.tracker.Subject.<init>(Subject.kt:185)
                 	at com.snowplowanalytics.core.tracker.ServiceProvider.makeSubject(ServiceProvider.kt:222)
                 	at com.snowplowanalytics.core.tracker.ServiceProvider.getOrMakeSubject(ServiceProvider.kt:181)
                 	at com.snowplowanalytics.core.tracker.ServiceProvider.makeTracker(ServiceProvider.kt:259)
                 	at com.snowplowanalytics.core.tracker.ServiceProvider.getOrMakeTracker(ServiceProvider.kt:189)
                 	at com.snowplowanalytics.core.tracker.ServiceProvider.<init>(ServiceProvider.kt:87)
                 	at com.snowplowanalytics.snowplow.Snowplow.createTracker(Snowplow.kt:275)
```

**To Reproduce**
```kotlin
  private val networkConfig = NetworkConfiguration(
    endpoint = "...",
    method = HttpMethod.POST
  )

  private val trackerConfig =
    TrackerConfiguration("my-namespace")
      .applicationContext(true)
      .base64encoding(false)
      .deepLinkContext(true)
      .diagnosticAutotracking(false)
      .exceptionAutotracking(true)
      .geoLocationContext(false)
      .installAutotracking(true)
      .lifecycleAutotracking(true)
      .platformContext(true)
      .platformContextProperties(
        listOf(
          PlatformContextProperty.AVAILABLE_STORAGE,
[...]

PR #683, opened 2024-04-12
Add Subject config flag to avoid WindowManager bug

mscwilson:
For issue #657.

This bug occurs during tracker initialisation when a) an Android application context from a dependency injection framework is passed in, and b) the Android StrictMode rule `detectIncorrectContextUse()` is set for the app.

That rule is violated when creating `Subject`, in `Subject.setDefaultScreenResolution()`, because it uses a deprecated method, via `WindowManager`, to fetch the screen size. The resolution set there in Subject is added to the payload as the "res" property, which eventually becomes the "dvce_screenwidth/height" columns.

The current preferred way is `DisplayManager` or from the `context.resources`. We already use the context Resources to get the screen resolution for the mobile/platform context entity.

Unfortunately, `WindowManager` gives a different answer from `DisplayManager`/`Resources`. The old way includes the menu bar at the top of the screen but that's not included anymore. Changing how Subject fetches the size is therefore a breaking change.

This PR adds a new flag to `SubjectConfiguration`. When set to true (false is default), the default Subject screen size is obtained from the context Resources and `WindowManager` is not accessed - meaning no StrictMode violation.

Currently, the flag is set up just like the other Subject properties. This doesn't seem necessary since it's only used in Subject initialisation. `SubjectController` doesn't really need it, but it inherits from `SubjectConfigurationInterface`. Anyway, it's Friday evening and this fix works.

### DIFF
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/Subject.kt
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/Subject.kt
@@ -179,13 +179,28 @@ class Subject(context: Context, config: SubjectConfigurationInterface?) {
             field = depth
             standardPairs[Parameters.COLOR_DEPTH] = depth.toString()
         }
+
+    var useContextResourcesScreenResolution: Boolean? = false
+        /**
+         * Whether to get the size from the context resources or not.
+         * By default this is false, the size is obtained from WindowManager.
+         *
+         * @param useContextResourcesScreenResolution
+         */
+        set(useContextResourcesScreenResolution) {
+            if (useContextResourcesScreenResolution == null) {
+                return
+            }
+
+            field = useContextResourcesScreenResolution
+        }
     
 
     init {
         setDefaultTimezone()
         setDefaultLanguage()
-        setDefaultScreenResolution(context)
-        
+        setDefaultScreenResolution(context, config?.useContextResourcesScreenResolution)
+
         if (config != null) {
             config.userId?.let { userId = it }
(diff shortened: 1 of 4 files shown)


snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/Subject.kt
```diff
@@ -179,13 +179,28 @@ class Subject(context: Context, config: SubjectConfigurationInterface?) {
             field = depth
             standardPairs[Parameters.COLOR_DEPTH] = depth.toString()
         }
+
+    var useContextResourcesScreenResolution: Boolean? = false
```
matus-tomlein, line 183: Why does this need to be optional?

matus-tomlein (approved):
LGTM! 👍

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/Subject.kt
mscwilson: It's set to null by default in SubjectConfig just to match the other properties

mscwilson: Do you recommend I make it non-optional?

matus-tomlein: I think that would be slightly better. It's a bit confusing what does the `null` state stand for and in the `TrackerConfiguration` we tend to use non-optional booleans for configuration like this.

PR #685, opened 2024-04-18
Release/6.0.3

mscwilson:
This patch release fixes two bugs.

The tracker will no longer crash if the app has a low version of the optional dependency `InstallReferrer`.

We've also added a new configuration flag, `useContextResourcesScreenResolution`, to fetch the default screen resolution - for the `dvce_screenwidth` and `dvce_screenheight` event properties - using a newer method.

NB: **the `dvce_screenwidth` and `dvce_screenheight` screen resolution values will be different when this flag is turned on**. 

```
val networkConfig = NetworkConfiguration("https://endpoint")
val subjectConfig = SubjectConfiguration().useContextResourcesScreenResolution(true)

Snowplow.createTracker(
    context, 
    "namespace",
    networkConfig,
    subjectConfig
)
```

When the flag is false (default), the default resolution is fetched using a deprecated method (`WindowManager`). This can cause StrictMode Violations under certain conditions. Newer methods of obtaining screen size don't include the menu bar.

The mobile context entity is not affected by the flag, as it already uses the newer Android API to obtain the screen resolution.

## Bug fixes

- Add Subject config flag to avoid WindowManager bug (#657)
- Catch more generic exceptions for InstallReferrer (#647)

---

BTW, the tests are failing for API 30 due to flakiness. This problem appeared quite recently and will be addressed in a future PR.


matus-tomlein (approved):
LGTM!

Issue #688, opened 2024-05-08
Network UserId Anonymization is delayed


JasminFrei:
**Describe the bug**
We heavily rely on the anonymization features the SDK offers. Our users can give or revoke their consent for tracking during a session in the app. If this happens, we will update the `tracker.userAnonymisation` and `tracker.emitter.serverAnonymisation` accordingly. What we can observe is the following behavior:
1. After setting `tracker.userAnonymisation` to `true`, a new session immediately gets created on the tracker without a userId, new Session Id and no previous session id. ✅
2. After setting `tracker.emitter.serverAnonymisation` we expect the new tracking requests to **immediately** have the `SP-Anonymous` header set to any new request so that no network user id gets tracked on server side. **Unfortunately this is not happening immediately, but only after a few requests/or some time** ❌

The issue is also present the other way round, so if we go from anonymous to non-anonymous, it takes a while until the header is not sent anymore. (But this direction has less legal implications)

**To Reproduce**
After initializing the trackers, change the server side anonymization and continue doing requests.

**Expected behavior**
After `tracker.emitter.serverAnonymisation` is set, I expect every new request to immediately send the `SP-Anonymous` header so that none of my tracked records for this new anonymous session to contain any user identifying data.

**Screenshots**

**Device informatoin (please complete the following information):**
- Emulator API 34
- OnePlus 6, Android 11 (much slower here until the header gets set/removed)

**Additional context**
This poses a legal issue as we can at least track the user for one session even if they think they're anonymous.

mscwilson:
Hi @JasminFrei, thanks for the report. We'll look into it

matus-tomlein:
This was fixed in version 6.0.4

PR #689, opened 2024-05-09
Set serverAnonymisation more quickly

mscwilson:
For issue #688 

Currently, when `serverAnonymisation` is set in the `Emitter`, the whole `OkHttpNetworkConnection` is recreated. This is slow, so there can be a delay in setting the anonymising `SP-Anonymous` header.

This change sets the `serverAnonymisation` directly in `OkHttpNetworkConnection`. Now the opposite is true - some events tracked just before the flag was set can end up with the `SP-Anonymous` header if they're still being processed when the change is made.

NB: the CI tests for API 30 are currently failing due to flakiness. This is a known problem and will be addressed in a future PR.

### DIFF
--- a/snowplow-tracker/src/main/java/com/snowplowanalytics/core/emitter/Emitter.kt
+++ b/snowplow-tracker/src/main/java/com/snowplowanalytics/core/emitter/Emitter.kt
@@ -197,9 +197,7 @@ class Emitter(
     var bufferOption: BufferOption = EmitterDefaults.bufferOption
         /**
          * Whether the buffer should send events instantly or after the buffer has reached
-         * its limit. By default, this is set to BufferOption Default.
-         *
-         * @param option Set the BufferOption enum to Instant to send events upon creation.
+         * its limit.
          */
         set(option) {
             if (!isRunning.get()) {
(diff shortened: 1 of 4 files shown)


snowplow-tracker/src/main/java/com/snowplowanalytics/core/emitter/Emitter.kt
```diff
@@ -303,19 +301,8 @@ class Emitter(
...
-                        .emitTimeout(it)
-                        .customPostPath(customPostPath)
-                        .client(client)
-                        .cookieJar(cookieJar)
-                        .serverAnonymisation(serverAnonymisation)
-                        .requestHeaders(requestHeaders)
-                        .build()
-                }
+            if (!isCustomNetworkConnection && builderFinished && networkConnection is OkHttpNetworkConnection) {
+                (networkConnection as OkHttpNetworkConnection).serverAnonymisation = true
```
matus-tomlein, line 305: Why is this always set to true? Shouldn't we set to the value we are given (the `serverAnonymisation`)?

mscwilson: 😫😳🤦🏼‍♀️

matus-tomlein (approved):
LGTM!

PR #691, opened 2024-06-04
Release/6.0.4

### LINKED ISSUE #687: Android InstallReferrerDetails: install_begin_timestamp & referrer_click_timestamp always wrong and install_begin_timestamp often missing
We came across that install_begin_timestamp and referrer_click_timestamp are always wrong, since the Snowplow SDK assumes that the Google Play installreferrer API response is in milliseconds but it is actually in seconds since epoch.

As you can see in this screenshot, I was able to proof this. I corrected the install_begin_timestamp through multiplying it by 1000 and got the correct timestamp. 
timestamp issue proof from BigQuery

**Source from the official API documentation regarding the response format**
- install_begin_timestamp: The timestamp, **in seconds**, of when a referrer click happened (both client- and server-side)
- referrer_click_timestamp: The timestamp, **in seconds**, of when an installation began (both client- and server-side)
https://developer.android.com/google/play/installreferrer

Related file:
https://github.com/snowplow/snowplow-android-tracker/blame/07536ea8db236e8f6efd74b07f3af901a136846e/snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/InstallReferrerDetails.kt#L70-L71

Further more, the install_begin_timestamp is missing for about 30% of installs (sample size ~145k installs on Android 8 and higher).We are running InstallReferrer package to v2.2 and Snowplow Tracker versions 5+
missing install_begin_timestamp

mscwilson:
This release fixes two bugs.

When calling `tracker.emitter.serverAnonymisation = true`, the flag is now set quickly, so that subsequent events tracked will immediately be anonymised.

For `ApplicationInstall` events, the timestamps in the `referrer_details` entity are now processed correctly. The timestamps provided by the Android InstallReferrer API are in seconds since epoch, but the tracker was processing them as milliseconds.

# Bug fixes
- Send correct install referrer timestamps (close #687)
- Set serverAnonymisation more quickly (close #689)

NB: the CI tests for API 30 are currently failing due to flakiness. This is a known problem and will be addressed in a future PR.


snowplow-demo-compose/src/main/java/com/snowplowanalytics/snowplowdemocompose/data/Tracking.kt
```diff
@@ -20,7 +20,7 @@ object Tracking {
     @Composable
     fun setup(namespace: String) : TrackerController {
         // Replace this collector endpoint with your own
-        val networkConfig = NetworkConfiguration("https://23a6-82-26-43-253.ngrok.io", HttpMethod.POST)
+        val networkConfig = NetworkConfiguration("http://192.168.0.20:9090", HttpMethod.POST)
```
matus-tomlein, line 23: This should probably be a placeholder [like we have in the RN tracker](https://github.com/snowplow/snowplow-react-native-tracker/blob/master/example/src/App.js#L34) as everyone will need to adjust to their own external IP.

matus-tomlein (approved):
LGTM!

PR #700, opened 2024-11-27
Add new WebView Interface

mscwilson:
This PR is part of our upgrade to the Snowplow WebView tracker. It allows the tracking of all web events, including certain base/primitive event properties.

Edit: test `tracksEventWithCorrectTracker()` is failing on Github, but it's fine locally 🫥


snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/TrackerWebViewInterfaceV2.kt
```diff
@@ -0,0 +1,140 @@
...
+import org.json.JSONObject
+import java.util.*
+
+/**
+ * JavaScript interface used to provide an API for tracking events from WebViews.
+ */
+class TrackerWebViewInterfaceV2 {
+    @JavascriptInterface
+    @Throws(JSONException::class)
+    fun trackWebViewEvent(
```
matus-tomlein, line 34: I'm slightly worried about this long list of parameters for 2 reasons:

1. It might cause breaking changes to introduce new parameters. What if we decide to also track `domain_sessionid` in the future? Will that make new WebView trackers tracking that extra property incompatible with older mobile trackers with WebViewInterface not having that extra parameter?
2. It's not a very nice API – I mean it's easy to make a mistake on the JS side and mess up the order of parameters being sent (there aren't named function params in JS).

My suggestion would be to consolidate the atomic properties into a JSON string and maybe have the following parameters:

1. selfDescribingEventData: String?
2. entities: String?
3. atomicProperties: String?

What do you think?

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/TrackerWebViewInterfaceV2.kt
```diff
@@ -0,0 +1,140 @@
...
+                if (tracker != null) {
+                    tracker.track(event)
+                } else {
+                    e(TAG, String.format("Tracker with namespace %s not found.", namespace))
+                }
+            }
+        }
+    }
+
+    @Throws(JSONException::class)
```
matus-tomlein, line 102: Nit: is this throwing a JSON exception? I don't see any JSON parsing in the function.

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/TrackerWebViewInterfaceV2Test.kt
```diff
@@ -0,0 +1,244 @@
...
+import org.junit.After
+import org.junit.Assert.*
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class TrackerWebViewInterfaceV2Test {
+    private var webInterface: TrackerWebViewInterfaceV2? = null
+    private var networkConnection = MockNetworkConnection(HttpMethod.GET, 200)
```
matus-tomlein, line 41: This is the old annoying way to test events that requires long sleeps for events to arrive in the requests. More recently we have been using the `EventSink` plugin where you can also check the payload properties without having to wait for the emitter to send the events (which we don't need to test in this test case).

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/TrackerEvent.kt
```diff
@@ -120,4 +133,18 @@ class TrackerEvent @JvmOverloads constructor(event: Event, state: TrackerStateSn
             Parameters.UNSTRUCTURED
         )
     }
+    
+    private fun wrapWebViewToPayload(toPayload: Payload, base64Encoded: Boolean) {
```
matus-tomlein, line 137: Very small nit: This is kind of a destructive function – if you call it once, it works fine, but if you call it twice, it will lose the self-describing data. It'd be better if it didn't call `remove` on `payload` and instead of filtered the properties of the `payload` when adding them over to `toPayload` like this:

```kt
        toPayload.addMap(payload.filterNot { it.key == Parameters.WEBVIEW_EVENT_DATA })
```

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/TrackerWebViewInterfaceV2Test.kt
matus-tomlein: On the other hand, that would allow you to test the wrapping of the self-describing data that happens in the `TrackerEvent`, but maybe that is worth a separate test file for that class and functionality?

snowplow-tracker/src/main/java/com/snowplowanalytics/core/tracker/TrackerWebViewInterfaceV2.kt
mscwilson: no named function params in JS 🫠

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/TrackerWebViewInterfaceV2Test.kt
mscwilson: I originally set up the tests using a filter plugin, but I changed it so that the `TrackerEvent` processing would be included.

Actually there's already a mix because the `addsEventNameAndSchemaForInspection()` test uses a filter (didn't realise we had a whole class for this, nice).

I could change it so the `tracksEventWithCorrectTracker()` test also uses afterTrack filter/`EventSink` - don't think it's worth separating out that first test (which relies on `TrackerEvent` stuff) into a new class. Unless you think it's worth it for neatness?

matus-tomlein: That makes sense! Yeah, if using the network connection is useful, we can keep that, don't have to complicate it. We can keep the first test using that.

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/TrackerWebViewInterfaceV2Test.kt
```diff
@@ -0,0 +1,195 @@
...
+            Thread.sleep(1000)
+            i++
+        }
+
+        assertEquals(1, networkConnection.countRequests())
+
+        val request = networkConnection.allRequests[0]
+        val payload = request.payload.map
+
+        assertEquals("ue", payload[Parameters.EVENT])
```
matus-tomlein, line 88: Can we add a test case for a structured or page view event that their event name is followed? I have an intuition that it will always be set to `ue` because of line 583 in `Tracker.kt`.

I might be wrong and the correct event name is assigned in the end, but a test case never hurts.

matus-tomlein (approved):
LGTM! 👍 

But can we keep this PR open until we have the WebView and JS tracker parts finalized? It's possible that we will find out that we need something extra and we haven't really tested the integration yet (since we don't have the other parts), so we may need some small changes/adjustments.

snowplow-tracker/src/androidTest/java/com/snowplowanalytics/snowplow/tracker/TrackerWebViewInterfaceV2Test.kt
mscwilson: it's assigned `ue` in Tracker in line 583, that method is called on L528, but then after that on L536 the `TrackerEvent` method `wrapPropertiesToPayload()` is called, where the webview atomic properties are added to the payload. So it ends up with the correct one.

I've updated the first test to use a different event name, to avoid adding another NetworkConnection test.

